PR #681, opened 2015-04-20
Add fix to translate table

itkg-nanne:


ApiBundle/Controller/TranslateController.php
```diff
@@ -26,10 +26,14 @@ class TranslateController extends BaseController
     public function translateAction(Request $request)
     {
         $displayedElements = $request->get('displayedElements');
+
+        $entityType = $this->container->camelize($request->get('entityType'));
+        $entityType = $this->container->underscore($entityType);
```
nicolasThal: Why do you have to first camelize then underscore the string?

itkg-nanne: I've got three kinds of entityType values : Camelize, underscore, lower case. If i underscore an underscored string, the method replace '_' by '.'. If I camelize a camelized string, there is no effect.

nicolasThal: All right.

With such a logic, I would prefer to put those line of code in a transformer, this would be easier to write a test for all cases :)

itkg-nanne:
done @nicolasThal

ApiBundle/Tests/Transformer/DisplayedElementCollectionTransformerTest.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public function getTranslatedDisplayedElements()
+    {
+        return array(
+            array('fakeEntityType', array('fake_column'), array('open_orchestra_backoffice.table.fake_entity_type.fake_column')),
+            array('fake_entity_type', array('fake_column'), array('open_orchestra_backoffice.table.fake_entity_type.fake_column')),
+            array('fake', array('fake_column'), array('open_orchestra_backoffice.table.fake.fake_column')),
+        );
+    }
+
```
nicolasThal: This empty line is too much :)

nicolasThal:
:+1: when you have suppressed the empty line :)

PR #683, opened 2015-04-21
Add role check on status changes

nicolasThal:


juchi:
The SecurityContextInterface (and therefore the security context) are deprecated and to be removed.
We should use the token_storage (for users) or the authorization_checker (for isGranted calls).

nicolasThal:
@juchi the modification has been made

itkg-bfouche:
:+1:

PR #694, opened 2015-04-23
Move user admin

nicolasThal:


UserAdminBundle/Controller/Admin/UserController.php
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @return Response
+     */
+    public function newAction(Request $request)
+    {
+        $user = new User();
+        $url = $this->generateUrl('open_orchestra_user_admin_new');
+        $event = UserEvents::USER_CREATE;
+        $form = $this->handleForm($request, 'registration_user', $user, $url, $event);
+        if ($form->isValid()) {
```
itkg-nanne: Maybe could you use handleForm here like in NodeController?

UserAdminBundle/Resources/config/leftpanel.yml
```diff
@@ -0,0 +1,10 @@
+services:
+    open_orchestra_user_admin.left_panel.user:
+        class: %open_orchestra_backoffice.left_panel.administration.class%
+        arguments:
+            - user
+            - ROLE_ACCESS_USER
+        calls:
+            - [ setTemplate, ['OpenOrchestraUserAdminBundle:AdministrationPanel:user.html.twig'] ]
+        tags:
+            - { name: open_orchestra_backoffice.left_panel.strategy }
```
itkg-nanne: eof

Backoffice/LeftPanel/Strategies/AdministrationPanelStrategy.php
```diff
@@ -79,4 +80,24 @@ public function getRole()
     {
         return $this->role;
     }
+
+    /**
+     * @return string
+     */
+    protected function getTemplate()
+    {
+        if ($this->template) {
```
itkg-ngilain: Can you check if the template exists before using it ? It will prevent crash from typo in service configuration and enhance stability :)

nicolasThal: I will do an issue for this ticket, I think there need to be a more global reflexion about that.

But I agree with you on this point

UserAdminBundle/Resources/config/leftpanel.yml
nicolasThal: correct in the https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/696 pr

UserAdminBundle/Controller/Admin/UserController.php
nicolasThal: I cannot, because we change the form if we are not creating the user

itkg-nanne: You could. After handleForm if count error is 0 then you redirect.

nicolasThal:
@itkg-nanne modification done

itkg-ngilain:
It's hard to have an opinion on this
so :+1: and let's see if there's a bug using it ^^'

PR #707, opened 2015-04-29
remove crop tab on non image media

itkg-ngilain:
https://trello.com/c/YAQFAdN4/839-1-etq-ubo-je-vois-l-onglet-crop-uniquement-pour-les-images

### DIFF
--- a/BackofficeBundle/Resources/views/BackOffice/Underscore/superboxView._tpl.twig
+++ b/BackofficeBundle/Resources/views/BackOffice/Underscore/superboxView._tpl.twig
@@ -13,7 +13,9 @@
     </ul>
 
     <div class="tab-content">
-        {% include 'OpenOrchestraBackofficeBundle:BackOffice:Underscore/Include/superboxView_crop._tpl.twig' %}
+        <% if (media.get('mime_type').indexOf('image') > -1) { %>
+            {% include 'OpenOrchestraBackofficeBundle:BackOffice:Underscore/Include/superboxView_crop._tpl.twig' %}
+        <% } %>
         {% include 'OpenOrchestraBackofficeBundle:BackOffice:Underscore/Include/superboxView_meta._tpl.twig' %}
     </div>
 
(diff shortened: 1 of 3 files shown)


BackofficeBundle/Resources/views/BackOffice/Underscore/superboxView._tpl.twig
```diff
@@ -3,54 +3,18 @@
...
-            <div id ="alternative-loader-container" class="form-group form-horizontal" style="display:none;">
-                <label class="control-label col-lg-3" /><div id="alternative-loader" />
-            </div>
-            <div class="media-override-format-form" style="display:none"></div>
-
-        </div>
-        <div class="tab-pane" id="tab-meta">
-            <div class="media_meta_form"></div>
-        </div>
+        {% include 'OpenOrchestraBackofficeBundle:BackOffice:Underscore/Include/superboxView_crop._tpl.twig' %}
```
nicolasThal: You could also do the same check for the file include

itkg-ngilain: No because the test isn't runned at the same time as the templates ares included :
. first twig generate the template including the js (underscore) test
. the generated template is sent to the backbone view
. then the js test is runned

It is so not possible at the twig time to know the media.
Another way to see it is that the twig is generic for all media and must therefore include both panes.

nicolasThal: Yes, I was aware :).

My point was only for the underscore test, so when you display the view, the whole crop part is not included

nicolasThal: It will be the same test as on line 6 for the panel

itkg-ngilain: Ok I see better whant you're talking about :)

itkg-ngilain:
@nicolasThal : update done

nicolasThal:
:+1:

PR #711, opened 2015-04-30
fix bug of field options of contentType

itkg-alavieille:


BackofficeBundle/Controller/ContentTypeController.php
```diff
@@ -77,22 +77,9 @@ public function newAction(Request $request)
...
-            $this->get('session')->getFlashBag()->add(
-                'success',
-                $this->get('translator')->trans('open_orchestra_backoffice.form.content_type.creation')
-            );
-
+        if (!$request->get('no_save')) {
+            $this->handleForm($form, $this->get('translator')->trans('open_orchestra_backoffice.form.content_type.creation'), $contentType);
             $this->dispatchEvent(ContentTypeEvents::CONTENT_TYPE_CREATE, new ContentTypeEvent($contentType));
-
-            return $this->redirect(
```
nicolasThal: it think that if the form is valid and there is no no_save parameter you should still do the redirect response

itkg-ngilain:
I let @itkg-alavieille answer to the @nicolasThal 's remark but it's ok for me if it works as described
:+1:

BackofficeBundle/Controller/ContentTypeController.php
itkg-alavieille: Why redirect to the edit form ?

nicolasThal: If you save the content-type, it is more userfriendly that you can directly edit it without going on an other page.

It is the workflow set on all the others cruds

itkg-alavieille: But this method is called with Ajax so if i make a redirect , it's not visible by user

itkg-alavieille:
@nicolasThal

BackofficeBundle/Controller/ContentTypeController.php
```diff
@@ -77,22 +77,14 @@ public function newAction(Request $request)
...
-                'success',
-                $this->get('translator')->trans('open_orchestra_backoffice.form.content_type.creation')
-            );
-
-            $this->dispatchEvent(ContentTypeEvents::CONTENT_TYPE_CREATE, new ContentTypeEvent($contentType));
-
-            return $this->redirect(
-                $this->generateUrl('open_orchestra_backoffice_content_type_form', array('contentTypeId' => $contentType->getContentTypeId()))
-            );
+                return $this->redirect($this->generateUrl('open_orchestra_backoffice_content_type_form', array(
```
nicolasThal: Befor the redirection, can you check if the contentType was valid and saved in the database (either you check the form, or if the mongo id ofthe contentType is set)

itkg-alavieille:
@nicolasThal

nicolasThal:
:+1:

PR #717, opened 2015-05-04
add full path for underscore template

itkg-nanne:


BackofficeBundle/Controller/UnderscoreTemplateController.php
```diff
@@ -11,23 +12,18 @@
...
+     * @param Request $request
      *
-     * @Config\Route("/underscore-template/show/{language}/{templateId}", name="open_orchestra_backoffice_underscore_template_show")
+     * @Config\Route("/underscore-template/show", name="open_orchestra_backoffice_underscore_template_show")
      * @Config\Method({"GET"})
      *
      * @return Response
      */
-    public function showAction($language, $templateId)
+    public function showAction(Request $request)
```
itkg-ngilain: You shall not remove language & template information from the route because they are discriminant information : 1 url = 1 unique resource.
Without those informations in the route, you can encounter problems with the navigator cache : the first ajax call will serve you a correct resource but the next calls will potentially use the navigator cache who was generated the first time for another couple language/template :/

itkg-ngilain:
:+1:

PR #718, opened 2015-05-05
Remove status from transverse

itkg-nanne:


BackofficeBundle/Manager/NodeManager.php
```diff
@@ -92,6 +97,20 @@ public function createNewLanguageNode(NodeInterface $node, $language)
     }
 
     /**
+     * @param NodeInterface $node
+     *
+     * @return StatusInterface
+     */
+    protected function getInitialStatus(NodeInterface $node)
+    {
+        if ($node->getNodeType() == NodeInterface::TYPE_GENERAL) {
```
nicolasThal: Why do you make a different test here and in the transformer : 
  - One for only node transverse
  - One for all general node

What about the other nodes?

BackofficeBundle/Tests/Manager/NodeManagerTest.php
```diff
@@ -226,7 +228,12 @@ public function generateConsistencyNode()
...
     {
-        $node = $this->manager->initializeNewNode();
+        $son = Phake::mock('OpenOrchestra\ModelInterface\Model\NodeInterface');
+        Phake::when($son)->getNodeType()->thenReturn('fakeType');
+        $fakeParentId = 'fakeParentId';
+        Phake::when($this->nodeRepository)->findOneByNodeIdAndLanguageAndVersionAndSiteId(Phake::anyParameters())->thenReturn($son);
+
+        $node = $this->manager->initializeNewNode($fakeParentId);
 
         $this->assertInstanceOf($this->nodeClass, $node);
```
nicolasThal: You could also add a check on the nodeType of the node returned!

nicolasThal:
I think that you should also add a test for the difference between the node transverse and the general nodes

BackofficeBundle/Manager/NodeManager.php
itkg-nanne: In fact TYPE_GENERAL is transverse. Others are TYPE_DEFAULT...

nicolasThal: Yes, but they can be TYPE_GENERAL nodes which are not transverse node.

The point of my remark was not about that but about the different test you where doing in the nodeTransformer and at this line

nicolasThal: https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/718/files#diff-51685802d3e8eddde0376c8bb9d9ae88R134 on this line

BackofficeBundle/Manager/NodeManager.php
```diff
@@ -92,6 +97,20 @@ public function createNewLanguageNode(NodeInterface $node, $language)
...
+            return $this->statusRepository->findOneByEditable();
```
nicolasThal: this method should not be used, use this method instead findOneByInitial

itkg-nanne: We're not sure that initial is not published on true. If transverse is initial and initial is published true thn we can't edit.

itkg-nanne:
c'est fait @nicolasThal

BackofficeBundle/Manager/NodeManager.php
```diff
@@ -92,6 +97,20 @@ public function createNewLanguageNode(NodeInterface $node, $language)
     }
 
     /**
+     * @param NodeInterface $node
+     *
+     * @return StatusInterface
+     */
+    protected function getInitialStatus(NodeInterface $node)
```
nicolasThal: I haven't seen it yet but this method is not consistent.

You cannot say getInitialStatus and then return only an editable status.

itkg-nanne: I use this beacause I need the initial status of the node, not the status which is flagged initial.
But I change it. @nicolasThal

nicolasThal:
:+1:

PR #736, opened 2015-05-12
revert changed tags

nicolasThal:


nicolasThal:
@itkg-nanne here is the fix

Issue #703, opened 2015-04-28
Typo in class name


juchi:
The class MediaChacheInvalidateSubscriber.php in MediaAdminBundle has a typo (Chache instead of Cache")

PR #740, opened 2015-05-13
rename chache to cache

nicolasThal:
should fix https://github.com/open-orchestra/open-orchestra-cms-bundle/issues/703 @juchi


itkg-ngilain:
:+1:

PR #746, opened 2015-05-18
 add namespace for ApiException in areaTransformerHttpException

itkg-alavieille:


ApiBundle/Exceptions/HttpException/AreaTransformerHttpException.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 namespace OpenOrchestra\ApiBundle\Exceptions\HttpException;
+use OpenOrchestra\BaseApi\Exceptions\ApiException;
```
nicolasThal: empty line before use

nicolasThal: And this is not the right class to use.

The ApiException to extend is the one in the OpenOrchestra\BaseApi\Exceptions\HttpException namespace

nicolasThal: I will fix it

PR #769, opened 2015-05-21
Translate block name in toolbar

itkg-ngilain:


nicolasThal:
Did you modify the test for your new check?

itkg-ngilain:
I tried but the actual tests doesn't allow it easily. I try with a new approach.

itkg-ngilain:
@nicolasThal : I have add a simple test on the translator call when it must be called.
I'd like to test the params when the ui_model transformer transforms the data, but as that parameter is a complex array and the html included in it can vary too, it's merely impossible.

nicolasThal:
I think that the check you made is ok. You can see it on scrutinizer : https://scrutinizer-ci.com/g/open-orchestra/open-orchestra-cms-bundle/inspections/eb5105fc-0ace-4789-968a-a2b3eb7679c4/code-structure/class/OpenOrchestra%5CApiBundle%5CTransformer%5CBlockTransformer

By clicking on the code coverage button, it will show you which line are tested

itkg-ngilain:
The test only tells us that the translator is called once, but not that the result is transmitted to the ui_model, and that's the important fact. But if it's ok for you, it's for me :)

nicolasThal:
If you want, you can add a parameter to the test to check the content of the ui_model

itkg-ngilain:
@nicolasThal : too hard because I have to test the hole array, including a second entry (the html) and that second entry is always set to 'ok' in our test, but the code can take it to special values such as 'icon'.

So again, if it's ok for you, let stay with this translator test

nicolasThal:
The tranlator test is ok for me then

nicolasThal:
:+1:

PR #781, opened 2015-05-27
sort content types in left panel

itkg-rboessel:


Backoffice/LeftPanel/Strategies/ContentTypeForContentPanelStrategy.php
```diff
@@ -30,7 +30,7 @@ public function __construct(ContentTypeRepositoryInterface $contentTypeRepositor
     public function show()
     {
         $contentTypes = $this->contentTypeRepository->findAllByDeletedInLastVersion();
-
+        ksort($contentTypes);
```
itkg-nanne: you should use the sort in mongo, not with php function.

nicolasThal: I agree with @itkg-nanne, the sort in php should never be used

PR #786, opened 2015-05-28
Site id in url

nicolasThal:


BackofficeBundle/Controller/HomepageController.php
```diff
@@ -15,11 +15,22 @@ class HomepageController extends Controller
      * Back Office Home Page
      *
      * @Config\Route("", name="homepage")
+     * @Config\Route("/homepage/{siteId}/{language}", name="homepage_with_site_id_and_language")
```
itkg-ngilain: If the siteId must be set in all BO url, then I think we must reorganize urls components : on a SEO point they should be organized from the high level info to the lower lvl, to get a coherent tree. So I'd use the {siteId} as a general prefix rather than a simple param. For this one it would give  /{siteId}/homepage/{language}

nicolasThal:
@itkg-ngilain update done

itkg-ngilain:
:+1:

PR #804, opened 2015-06-04
Data table pagination ajax

itkg-alavieille:
in relation with : 
https://github.com/open-orchestra/open-orchestra-base-api-bundle/pull/17
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/315
https://github.com/open-orchestra/open-orchestra-model-interface/pull/71
https://github.com/open-orchestra/open-orchestra-user-bundle/pull/56
https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/pull/6


ApiBundle/Controller/ContentController.php
```diff
@@ -101,10 +101,32 @@ public function showOrCreateAction(Request $request, $contentId)
     public function listAction(Request $request)
     {
         $contentType = $request->get('content_type');
+        $columns = $request->get('columns');
```
nicolasThal: isn't there a way to factorize the creation of all those parameters?

ApiBundle/Facade/PaginateCollectionFacade.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @Serializer\Type("integer")
+     * @Serializer\SerializedName("recordsFiltered")
+     */
+    public $recordsFiltered;
+
+    /**
+     * @param $recordsTotal
+     */
+    public function setRecordsTotal($recordsTotal)
```
nicolasThal: as the properties are public and they are integer, you do not need a setter to set them

LogBundle/Facade/LogCollectionFacade.php
```diff
@@ -36,4 +48,20 @@ public function getLogs()
     {
         return $this->logs;
     }
+
+    /**
+     * @param $recordsTotal
+     */
+    public function setRecordsTotal($recordsTotal)
```
nicolasThal: do not put setters

nicolasThal:
there is a lot of code duplication.

I think that there would be a way to factorize it

itkg-alavieille:
@nicolasThal

ApiBundle/Controller/ContentController.php
nicolasThal: did you forgot to refactor?

nicolasThal:
you still can suppress the setter in the facade and directly give the value to the property : $facade->count = $count

ApiBundle/Controller/ContentController.php
itkg-alavieille: The contentController, ContentController and SiteController use the different method in repository which add additional filters so i can't use the trait HandleRequestDataTable

nicolasThal: Are you sure?

You could give as a parameter the name of the method in the repository for example

itkg-alavieille: Yes but this method take orther parameter. for instance in the contentController the method for pagination take a parameter contentType.

nicolasThal: all right.

So in this case, I would only suggest to refactor the extraction of all the parameter of the request in an external method

itkg-alavieille:
@nicolasThal

nicolasThal:
:+1:

PR #808, opened 2015-06-05
add cache in dataTable and direct access

itkg-alavieille:


BackofficeBundle/Resources/public/js/backboneRouter.js
```diff
@@ -12,7 +12,7 @@ var OrchestraBORouter = Backbone.Router.extend({
     'node/show/:nodeId/:language': 'showNodeWithLanguage',
     'node/show/:nodeId/:language/:version': 'showNodeWithLanguageAndVersion',
     'template/show/:templateId': 'showTemplate',
-    ':entityType/list': 'listEntities',
+    ':entityType/list(/:page)': 'listEntities',
```
nicolasThal: Nice work, I think that we could add a refactoring ticket for the route on the nodes with the language and the version optional

itkg-alavieille: yes i add a ticket

nicolasThal:
:+1:

PR #831, opened 2015-06-11
fix duplicate entry in grunt symlink task

itkg-ngilain:
https://github.com/open-orchestra/open-orchestra/issues/459


nicolasThal:
:+1:

PR #832, opened 2015-06-11
add site linkable behaviour

nicolasThal:


ApiBundle/Controller/ContentController.php
```diff
@@ -113,7 +114,7 @@ public function listAction(Request $request)
         );
```
itkg-alavieille: for filter the contents according they are linked to a site, should update the mapping with 
'site_linked' => array('key' => 'siteLinked', 'type' => 'boolean'

ApiBundle/Controller/ContentController.php
```diff
@@ -113,7 +114,7 @@ public function listAction(Request $request)
...
 
         $repository =  $this->get('open_orchestra_model.repository.content');
-        $contentCollection = $repository->findByContentTypeInLastVersionForPaginateAndSearch($contentType, $columnsNameToEntityAttribute, $columns, $search, $order, $skip, $limit);
+        $contentCollection = $repository->findByContentTypeInLastVersionForPaginateAndSearchAndSiteId($contentType, $columnsNameToEntityAttribute, $columns, $search, $siteId, $order, $skip, $limit);
```
itkg-ngilain: This naming is becoming totallly absurd :s
@itkg-alavieille dind'nt want to enumerate all the params so he suffixed it with 'ForPaginateAndSearch' instead. Now there's a random param added to the end ('AndSiteId'). Where is the consistency to talk about the SiteId and not about the other params ? (Without speaking of the ridiculous naming the method is going to have now :p)

nicolasThal: Yes, but contrary to all the search parameters, which are given by  the DataTable request, the siteId is an internal parameter, that's why I added it in the method name

ApiBundle/Controller/ContentController.php
nicolasThal: discussed with @itkg-alavieille and this cannot be done

ApiBundle/Controller/ContentController.php
itkg-ngilain: I don't make this a 'PR-stopper' as we should have a general brainstorming on the naming of the repo methods.

itkg-ngilain:
:+1:

PR #834, opened 2015-06-12
Load user language

itkg-nanne:


Backoffice/Context/ContextManager.php
```diff
@@ -20,8 +19,6 @@ class ContextManager implements CurrentSiteIdInterface
     protected $siteId;
     protected $session;
     protected $tokenStorage;
-    protected $defaultLocale;
```
itkg-alavieille: you removed attribute defaultLocale but in the constructor you set it and use it in the method getCurrentLocale

Backoffice/Context/ContextManager.php
```diff
@@ -32,17 +29,11 @@ class ContextManager implements CurrentSiteIdInterface
      * @param TokenStorageInterface   $tokenStorage
      * @param string                  $defaultLocale
      */
-    public function __construct(Session $session, SiteRepositoryInterface $siteRepository, TokenStorageInterface $tokenStorage, $defaultLocale = 'en')
+    public function __construct(Session $session, TokenStorageInterface $tokenStorage, $defaultLocale)
```
itkg-alavieille: removes the comment parameter "param SiteRepositoryInterface $siteRepository"

Backoffice/Context/ContextManager.php
```diff
@@ -52,7 +43,19 @@ public function __construct(Session $session, SiteRepositoryInterface $siteRepos
      */
     public function getCurrentLocale()
     {
-        return $this->session->get(self::KEY_LOCALE);
+        $currentLanguage = $this->session->get(self::KEY_LOCALE);
+
+        if (!$currentLanguage) {
+            $currentLanguage = $this->defaultLocale;
+            $token = $this->tokenStorage->getToken();
+            if ($token && ($user = $token->getUser()) instanceof GroupableInterface) {
```
nicolasThal: Why GroupableInterface to check if there is an user?

BackofficeBundle/Controller/ContentController.php
```diff
@@ -28,7 +28,7 @@ public function formAction(Request $request, $contentId)
     {
         $language = $request->get(
             'language',
-            $this->get('open_orchestra_backoffice.context_manager')->getDefaultLocale()
+            $this->get('open_orchestra_backoffice.context_manager')->getCurrentSiteDefaultLanguage()
```
nicolasThal: I don't agree with this modification?

What does @nbouquet says?

BackofficeBundle/Manager/ContentManager.php
```diff
@@ -48,7 +48,7 @@ public function initializeNewContent($contentType)
         $contentClass = $this->contentClass;
         /** @var ContentInterface $content */
         $content = new $contentClass();
-        $content->setLanguage($this->contextManager->getDefaultLocale());
+        $content->setLanguage($this->contextManager->getCurrentLocale());
```
nicolasThal: I don't agree with this modification

nicolasThal:
It is the first time I will do it but I will give :-1: to this PR because : 

 - It mixes too many things
 - Why do you need to get Twig to get the availables languages?
 - Why do you modify the way content are created ?

I think that this pr should only be about the BO language linked to the user, you should split it for the others behaviour

itkg-nanne:
I rollback on content creation.
I need twig because available languages for BO are in twig.globals.available_languages (see TwigAvailableLanguagesCompilerPass in open-orchestra-base-bundle).
I need Form user fully functional to test this features. So I mixed it with a fix on formPanel and on userType. @nicolasThal

UserAdminBundle/Form/Type/UserType.php
```diff
@@ -13,13 +14,21 @@
 class UserType extends AbstractType
 {
     protected $class;
+    protected $languages;
 
     /**
      * @param string $class
      */
-    public function __construct($class)
+    public function __construct($class, Twig_Environment $twig)
```
nicolasThal: you should use this parameter : `open_orchestra_base.languages.availables` of the symfony kernel and pass it directly to the class, this will be a more cleaner way that to get the `Twig_Environment`

itkg-nanne: And if this parameter is empty, twig.globals.available_laguages will have fr and en as default value and none will be selectable in user form?

nicolasThal: In this file `OpenOrchestraBaseExtension` The languages 'fr' and 'en' are always set.

In the compiler pass, we should suppress the default values I think

itkg-nanne:
test functional ok, @nicolasThal

UserAdminBundle/Form/Type/UserType.php
```diff
@@ -13,13 +14,19 @@
...
 {
     protected $class;
+    protected $languages;
 
     /**
      * @param string $class
+     * @param array  $availableLanguages
      */
-    public function __construct($class)
+    public function __construct($class, $availableLanguages)
```
nicolasThal: in the signature, you can make ($class, array $availableLanguages)

nicolasThal:
:+1: with the signature modification

Issue #837, opened 2015-06-12
Error when using "back to list" on new content creation


juchi:
I try to create a new content by clicking the "Add" button. I don't save my form and just click "back to list" on the subsequent page, then the "loading" spinner is displayed and the page doesn't load.
In the console there is a JS error "link.data(...) is undefined".

itkg-alavieille:
This pr will fix this https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/839

PR #839, opened 2015-06-15
fix link back_to_list in fullpageFormView

itkg-alavieille:
fix this issue https://github.com/open-orchestra/open-orchestra-cms-bundle/issues/837


nicolasThal:
:+1:

PR #847, opened 2015-06-16
replace span by div and update css

itkg-alavieille:


juchi:
Wouldn't this be a good time to rename this class "display-block-icon-font" to something less CSSized ?

itkg-alavieille:
what do you mean by CSSized

juchi:
Prefixing the class with "display-block" infers the use of display:block (which is not good practice anyway), and a div is by default displayed as a block, so i guess it's no more needed (the prefix and the properties if they exist)

itkg-alavieille:
i think the class does not refer to the css property but at blocks like (configurable content, content list, etc).  I will change it  just by block-icon-font. this will be more clear

itkg-alavieille:
@juchi

juchi:
:+1:

PR #849, opened 2015-06-17
allow multiple tinyMce and activate record

itkg-nanne:


MediaAdminBundle/Resources/public/coffee/bindExtendView.coffee
```diff
@@ -1,20 +1,18 @@
...
+        $.extend true, view, extendView['orchestraMediaAbstractType'], extendView['orchestraWysiwygType']
+        target = editor.id + '_modal'
+        $('#' + editor.editorContainer.id + ' .mce-btn[aria-label="mediamanager"] button').data
+          target: target
+          input: editor.id
+          url: $('#' + target).data('url')
+        view.delegateEvents()
+        return
+      return
+  return
```
itkg-alavieille: eof

BackofficeBundle/Resources/views/layout.html.twig
```diff
@@ -84,6 +84,6 @@
             pageSetUp();
         });
     </script>
-
+    {{ tinymce_init() }}
```
itkg-alavieille: Why init tinymce in all pages, you can't init it just in the page with form which uses tinymce ?

itkg-nanne: I override script.html.twig by extending bundle. So tinmce_init() will just load js library fo tiny, but not really init it.  initTinyMCE(); is in widgetChannel.coffee.

BackofficeBundle/Resources/views/layout.html.twig
```diff
@@ -84,6 +84,6 @@
             pageSetUp();
         });
     </script>
-
+    {{ orchestra_tinymce_init() }}
```
juchi: I think (not sure) that the only instance of tinymce_init() is located in form_div_layout.html.twig, so it would be better to update the twig extension rather than creating a new one.

juchi:
I'm not sure to see the point of including tinymce on the layout. There is already quite an issue about performances and tinymce is needed on very few pages to my knowledge.
Why not loading it on the fly when it's needed ?

nicolasThal:
By doing this, the goal is to load the js file for tinymce only once.

After the tinymce is launched directly only when it is needed

juchi:
I get this. My point was that it would be more efficient and long-term to develop a small component that allows to load any JS/CSS file dynamically, and to check that it is loaded only once. This would avoid half-fixing the tinymce case and be extensible to all kinds of scripts.

itkg-nanne:
done @nicolasThal

nicolasThal:
:+1:

PR #860, opened 2015-06-18
refacto name backbone model

itkg-alavieille:


BackofficeBundle/Resources/public/coffee/modelBackbone.coffee
```diff
@@ -1,27 +1,28 @@
-Site = Backbone.Model.extend({})
-Block = Backbone.Model.extend({})
-BlockCollection = Backbone.Collection.extend(model: Block)
+SiteModel = Backbone.Model.extend({})
+
+BlockModel = Backbone.Model.extend({})
+BlockCollection = Backbone.Collection.extend(model: BlockModel)
```
itkg-ngilain: I'd rather use the name BlocksCollection, because it's a collections of blocks

BackofficeBundle/Resources/public/coffee/modelBackbone.coffee
```diff
@@ -1,27 +1,28 @@
...
+
 Area = Backbone.Model.extend(
   blocks: BlockCollection
   areas: AreaCollection
 )
-AreaCollection = Backbone.Collection.extend(model: Block)
-Node = Backbone.Model.extend(
+AreaCollection = Backbone.Collection.extend(model: BlockModel)
```
itkg-ngilain: Same here with AreasCollection and the other collections

itkg-ngilain:
Except the naming problem, it's a :+1: for me if the other dev agree with you

BackofficeBundle/Resources/public/coffee/modelBackbone.coffee
nicolasThal: I think that we could use the naming the same way we do for the facade. We only use singular : `NodeCollection` `SiteCollection` ...

nicolasThal:
:+1:

juchi:
I hope this will make it to the ChangeLog :)

PR #870, opened 2015-06-19
Refresh nav on content type creation

itkg-ngilain:


BackofficeBundle/Resources/public/coffee/FullPageFormView.coffee
```diff
@@ -36,6 +36,7 @@ FullPageFormView = OrchestraView.extend(
     $("form", @$el).ajaxSubmit
       context: button: $(".submit_form",event.target).parent()
       success: (response) ->
+        displayMenu() if $(event.target).attr('name') is 'content_type'
```
itkg-nanne: I would prefer to have a dedicated view for content_type which will be loaded by appConfigurator. This view could override FullPageFormView by adding displayMenu() in initialize.

itkg-ngilain:
@itkg-nanne : update done, but I don't like the overload of the initialize function

itkg-nanne:
An other way could be to bind on ready event after render. In widgetChannel, if view.options.submitted and view.options.afterSubmitted then view.options.afterSubmitted(). You keep same override logic but just with an afterSubmitted method. Could be used for folder, node and template. right?

itkg-nanne:
In case of new ContentType, will it open navigation on contentType created?

itkg-ngilain:
@itkg-nanne : update done introducing an onOrchestraViewReady and onViewReady functions

itkg-nanne:
:+1:  after rebase

PR #871, opened 2015-06-22
allow multiple tinyMce open without submit

itkg-nanne:


MediaAdminBundle/Resources/public/coffee/bindExtendView.coffee
```diff
@@ -1,18 +1,22 @@
...
-          target: target
-          input: editor.id
-          url: $('#' + target).data('url')
-        view.delegateEvents()
-        return
-      return
   return
+
+#ACTIVATE TINYMCE
+callback_tinymce_init = null
```
nicolasThal: Does this part needs to be in this file? or can it be split into another file?

juchi: The naming of the variable is not consistent with the rest of the project (CamelCase ?)
The variable should be assigned with a function, not null. This function could throw an exception or a console.error to inform that it has not been initiliazed.

itkg-nanne: Could be split, but launched by the ready bind. That's why i choose to put this here. As you wish...

juchi:
As a side note, I'd advise against global variables and functions.

MediaAdminBundle/Resources/public/coffee/bindExtendView.coffee
itkg-nanne: @juchi , callback_tinymce_init comes from tinyMce. We can't change it. Automatically called when tinyMce loaded.

itkg-nanne:
@juchi : https://github.com/stfalcon/TinymceBundle#init-event

juchi:
Ok, my bad.

PR #872, opened 2015-06-22
refacto api bundle

itkg-alavieille:


ApiBundle/Controller/AreaController.php
```diff
@@ -154,16 +154,19 @@ public function deleteAreaFromTemplateAction($areaId, $templateId)
...
     {
+        $areaContainer = null;
+
         if ($nodeId) {
             $nodeRepository = $this->get('open_orchestra_model.repository.node');
             $node = $nodeRepository->find($nodeId);
             $areaContainer = $nodeRepository->findAreaByAreaId($node, $parentAreaId);
-        }
-        if ($templateId && is_null($nodeId)) {
+        } else if ($templateId && is_null($nodeId)) {
```
nicolasThal: the second part of the test is not usefull I think.

If nodeId is not null, it will go directly in the if, and not in the elseif

ApiBundle/Controller/ControllerTrait/HandleRequestDataTable.php
```diff
@@ -5,6 +5,7 @@
 use Doctrine\ODM\MongoDB\DocumentRepository;
 use OpenOrchestra\BaseApi\Transformer\TransformerInterface;
 use Symfony\Component\HttpFoundation\Request;
+use OpenOrchestra\BaseApi\Facade\FacadeInterface;
```
nicolasThal: As this is a trait, I don't think that the use are useFull, especially for hte comment

ApiBundle/Controller/ControllerTrait/ListStatus.php
```diff
@@ -1,7 +1,8 @@
 <?php
 
-namespace OpenOrchestra\ApiBundle\Controller;
+namespace OpenOrchestra\ApiBundle\Controller\ControllerTrait;
```
nicolasThal: You should add this movement into the changelog

ApiBundle/Tests/Transformer/ContentTransformerTest.php
```diff
@@ -85,6 +84,10 @@ public function testTransform()
 
     /**
      * test reverseTransform
+     * @param FacadeInterface  $facade
```
nicolasThal: you should either : 
 - let one empty line between the comment and the first @param
 - suppress the comment

ApiBundle/Transformer/ApiClientCollectionTransformer.php
```diff
@@ -2,7 +2,7 @@
 
 namespace OpenOrchestra\ApiBundle\Transformer;
 
-use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\ArrayCollection;
```
nicolasThal: `ArrayCollection` extends `Collection`, but you could also get a PersistentCollection from the database.

I would prefer to let only the `Collection`, this allows more modification.

ApiBundle/Transformer/ApiClientCollectionTransformer.php
```diff
@@ -13,15 +13,15 @@
 class ApiClientCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param Collection $mixed
+     * @param ArrayCollection $apiClientCollection
      *
      * @return FacadeInterface
      */
-    public function transform($mixed)
+    public function transform($apiClientCollection)
```
nicolasThal: you should let $mixed as the parameter name, this way it will be consistent with the interface

ApiBundle/Transformer/ApiClientTransformer.php
```diff
@@ -13,24 +13,24 @@
 class ApiClientTransformer extends AbstractTransformer
 {
     /**
-     * @param ApiClient $mixed
+     * @param ApiClientInterface $apiClient
      *
      * @return FacadeInterface
      */
-    public function transform($mixed)
+    public function transform($apiClient)
```
nicolasThal: let $mixed for the parameter name

ApiBundle/Transformer/NodeCollectionTransformer.php
```diff
@@ -14,15 +14,15 @@
 class NodeCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param ArrayCollection $mixed
+     * @param ArrayCollection $nodeCollection
```
nicolasThal: You should also put `Collection` instead of `ArrayCollection`

ApiBundle/Transformer/RoleCollectionTransformer.php
```diff
@@ -13,15 +13,15 @@
 class RoleCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param ArrayCollection $mixed
+     * @param ArrayCollection $roleCollection
```
nicolasThal: `Collection` instead of `ArrayCollection`

ApiBundle/Transformer/SiteCollectionTransformer.php
```diff
@@ -13,15 +13,15 @@
 class SiteCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param ArrayCollection $mixed
+     * @param ArrayCollection $siteCollection
```
nicolasThal: same

ApiBundle/Transformer/StatusCollectionTransformer.php
```diff
@@ -13,16 +14,16 @@
 class StatusCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param ArrayCollection $mixed
-     * @param StatusInterface $currentStatus
+     * @param ArrayCollection      $statusCollection
```
nicolasThal: same

ApiBundle/Transformer/ThemeCollectionTransformer.php
```diff
@@ -13,15 +13,15 @@
 class ThemeCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param ArrayCollection $mixed
+     * @param ArrayCollection $themeCollection
```
nicolasThal: same

nicolasThal:
imho, I would rather let the first parameter given to the transform method named `$mixed`, this way, it will be more consistent with the `TransformerInterface`

ApiBundle/Controller/ControllerTrait/HandleRequestDataTable.php
itkg-alavieille: the publisher doesn't know the type of the return value and so the functionality (detect deprecated, type, etc.) provided by is lost

itkg-alavieille:
Yes but on the other side mixed isn't very explicit

itkg-alavieille:
@nicolasThal

ApiBundle/Controller/DeletedController.php
```diff
@@ -27,7 +26,7 @@ class DeletedController extends BaseController
      */
     public function listAction()
     {
-        /** @var Collection $nodes */
+        /** @var Array $nodes */
```
nicolasThal: you should put `array` instead of `Array` and put the comment line just above the $nodes line definition

nicolasThal:
:+1: with the small modification

ApiBundle/Transformer/ApiClientCollectionTransformer.php
```diff
@@ -13,15 +13,15 @@
 class ApiClientCollectionTransformer extends AbstractTransformer
 {
     /**
-     * @param Collection $mixed
+     * @param Collection $apiClientCollection
      *
      * @return FacadeInterface
      */
-    public function transform($mixed)
+    public function transform($apiClientCollection)
```
itkg-ngilain: Can we use this refacto to enhance the stability ? As the param ($mixed) has no explicit type in the interface, we should at least check here that the param has the expected type in that context and raise an error if not.
This should be done for all transformers.

itkg-alavieille:
@nicolasThal @itkg-ngilain

ApiBundle/Exceptions/HttpException/TransformParameterTypeHttpTypeHttpException.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace OpenOrchestra\ApiBundle\Exceptions;
+
+/**
+ * Class TransformerParameterException
+ */
+class TransformerParameterTypeHttpException extends \Exception
```
nicolasThal: This exception should extent `OpenOrchestra\BaseApiBundle\Exception\ApiException`

ApiBundle/Exceptions/HttpException/TransformParameterTypeHttpTypeHttpException.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    const DEVELOPER_MESSAGE  = 'open_orchestra_api.transformer.type';
```
nicolasThal: As this exception should not be thrown on user (and so serialized), I do not think that you need all those constant

nicolasThal:
:+1:

PR #873, opened 2015-06-22
fix overlay and no image

itkg-nanne:


juchi:
why prefixing a css class with "div" ?
css classes should use lisp-case

itkg-nanne:
@juchi , class come from smartadmin : https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/BackofficeBundle/Resources/public/smartadmin/css/smartadmin-production-plugins.css

itkg-rboessel:
:+1:

PR #875, opened 2015-06-22
fix form submission in FullPageFormView

juchi:
If the FullPageFormView has multiple form in it, it will always submit the first one no matter what.
As we have direct access to the actually submitted form, I use it in the jQuery selector.


itkg-nanne:
FullPagePanelView.coffe do it. Why not using tabs when you've got several forms?

nicolasThal:
What will happen when the submit button is outsite the form balise

juchi:
@nicolasThal Nothing changes as the target is the form itself.
@itkg-nanne I actually implemented it before this feature exists, maybe I should update. Anyway, it seems cleaner to update it there too, because you could have several forms on a given panel.

itkg-nanne:
ok :+1:

PR #882, opened 2015-06-24
update deprecated_form_type

itkg-alavieille:


BackofficeBundle/Form/Type/AreaType.php
```diff
@@ -60,15 +64,15 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             'label' => 'open_orchestra_backoffice.form.area.bo_percent'
         ));
         if(!array_key_exists('disabled', $options) || $options['disabled'] === false){
-            $builder->addEventSubscriber(new AreaCollectionSubscriber($this->areaClass));
+            $builder->addEventSubscriber(new AreaCollectionSubscriber($this->areaClass, $this->translator));
```
nicolasThal: As there are a lot of dependancies for this service, could we declare this subscriber as a service and pass it directly to the `AreaType`

BackofficeBundle/Form/Type/NodeType.php
```diff
@@ -119,23 +123,21 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         if(!array_key_exists('disabled', $options) || $options['disabled'] === false){
             $builder->addEventSubscriber(new NodeChoiceSubscriber($this->nodeManager));
             $builder->addEventSubscriber(new TemplateChoiceSubscriber($this->templateRepository));
-            $builder->addEventSubscriber(new AreaCollectionSubscriber($this->areaClass));
+            $builder->addEventSubscriber(new AreaCollectionSubscriber($this->areaClass, $this->translator));
```
nicolasThal: you could use the same service here :)

BackofficeBundle/Form/Type/AreaType.php
itkg-ngilain: It's not the only case where we use a new instead of a service. I've made a ticket for that as it's a general problem (https://trello.com/c/VEyutJDJ/1101-plusieurs-objets-sont-crees-avec-des-news-plutot-qu-en-passant-par-un-service-generaliser-l-appel-aux-services-pourra-necessiter)
As this PR is already consequent and there's some other cases of this, I suggest to put these refacto on a separated PR.

nicolasThal: works for me :)

itkg-alavieille:
so it's goot for the PR @nicolasThal @itkg-ngilain

nicolasThal:
:+1:

PR #892, opened 2015-06-29
Activate html5 validation for hidden

itkg-nanne:


BackofficeBundle/Resources/public/coffee/orchestraLib.coffee
```diff
@@ -164,13 +164,7 @@ activateTinyMce = (view, textarea) ->
...
         if $(this).data('required')
           $(this).attr('required', 'required')
-      textarea.css(
-         display: 'block'
-         position: 'absolute'
-         width: '1px'
-         height: '1px'
-         top: '0px'
-         zIndex: '1')
+      textarea.css({display: 'block'}).addClass('focusable')
```
itkg-ngilain: Can we remove css diplay from js to put it in css files using a new class instead ?

itkg-nanne: @itkg-ngilain , no because "tinyMce lib" put "display:none" in style. Only way to override it, is to put it in style too.

itkg-ngilain: Yes but can't you do this instead ? 

    textarea.addClass('displayBlock')

and in a stylesheet

    .displayBlock {display: block;}

This way there's no more css in our js

itkg-ngilain:
:+1:

PR #900, opened 2015-06-30
fix issue on content type reload

itkg-nanne:


BackofficeBundle/Resources/public/js/backboneRouter.js
```diff
@@ -166,20 +166,17 @@ var OrchestraBORouter = Backbone.Router.extend({
 
   generateUrl: function(routeName, paramsObject)
   {
+    var optionalParam = /\(([^)]*):([^)]*)\)/g;
```
nicolasThal: you should precise the _ in the regex

itkg-nanne: Why?

nicolasThal: So your regex would be more  linked to the pattern you have put `/\(([^)]*_):([^)]*)\)/g`

itkg-nanne: the pattern is an arbitrary choice. Could be language.fr or languagefr or anything else. But in order to have no conflict, I add '_'. Here we are in remplacement process, so separator character has no importance. We could choose to change separator in route, if we got conflict (i.e. language_fr is an entity) without change resolve routing process. @nicolasThal

nicolasThal:
:+1:

PR #909, opened 2015-07-01
add debug bar for call ajax

itkg-alavieille:


nicolasThal:
does it also appear for the api call

itkg-alavieille:
if you call directly the api it doesn't appear but if you call api from a page which has a debug bar it will be updated. In fact the debug bar does appears more than before, if in page there is a debug bar, it is updated for each ajax call.

nicolasThal:
:+1:

PR #912, opened 2015-07-02
Create base api model bundle

nicolasThal:


itkg-nanne:
:+1: with test

nicolasThal:
tests will be ok when the pr in media-bundle is merged :)

PR #915, opened 2015-07-02
Small modifications

nicolasThal:


BackofficeBundle/Form/Type/TemplateType.php
```diff
@@ -40,9 +40,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ->add('name', 'text', array(
                 'label' => 'open_orchestra_backoffice.form.template.name',
             ))
-            ->add('language', 'orchestra_language', array(
```
itkg-ngilain: @nbouquet wanted in a mid-term period to be able to put "default" blocks such has header/footer on templates in order to avoid their creations on each node. Removing the language from the templates breaks this feature. Is this refacto ok with that ?

nbouquet: I can confirm that we will put default or even transversal blocks on templates.

PR #917, opened 2015-07-02
Fix media folder creation

itkg-ngilain:


MediaAdminBundle/Form/Type/OrchestraSiteForFolderChoiceType.php
```diff
@@ -0,0 +1,99 @@
...
+                'embed' => false,
+                'choices' => $this->getChoices()
+            )
+        );
+    }
+
+    /**
+     * @return array
+     */
+    public function getChoices()
```
nicolasThal: this method should be protected

MediaAdminBundle/Tests/Form/Type/OrchestraSiteForFolderChoiceTypeTest.php
```diff
@@ -0,0 +1,143 @@
...
+                'choices' => $this->choiceList
+        ));
+    }
+
+    /**
+     * Test Choices list creation 
+     */
+    public function testGetChoices()
+    {
+        $this->assertSame($this->choiceList, $this->form->getChoices());
```
nicolasThal: the test of this method is done it the testConfigureOptions

MediaAdminBundle/Form/Type/OrchestraSiteForFolderChoiceType.php
itkg-ngilain: I agree with you, it was protected until I made unit tests. I need to test it as it contains the specific logic of the formtype, but being protected it can't :/
Do you have an alternative ?

MediaAdminBundle/Tests/Form/Type/OrchestraSiteForFolderChoiceTypeTest.php
itkg-ngilain: It's no more a unit test if you're testing a method not directly but by testing the result of another which may probably (or not) use it. No ?

itkg-ngilain:
@nicolasThal : update done

nicolasThal:
:+1:

PR #932, opened 2015-07-09
Fix media url problem in tinymce

itkg-ngilain:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/932
https://github.com/open-orchestra/open-orchestra-display-bundle/pull/157
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/368


itkg-nanne:
:+1:

PR #936, opened 2015-07-10
activate select2 in not backbone view context

itkg-nanne:


MediaAdminBundle/Resources/public/coffee/SuperboxView.coffee
```diff
@@ -94,6 +94,8 @@ SuperboxView = OrchestraView.extend(
       success: (response) ->
         $('#selector-loader-container').hide()
         $('.media_crop_form', currentView.$el).html response
+        if (form = $('.media_crop_form form', currentView.$el)) && form.length > 0
```
nicolasThal: Would it be a good thing to refactor those two methods?

itkg-nanne: clearly yes! In a first step, we should use the tab logic as in user's form. In a second step, we should be able to delete the two calls to hide method in SetupCropForm and factorize displayLoader. Then, third step, activateForm for each tab. Create ticket @nicolasThal

nicolasThal:
:+1:

PR #938, opened 2015-07-10
add alias id and node id to preview url

nicolasThal:


nicolasThal:
https://github.com/open-orchestra/open-orchestra-base-bundle/pull/32
https://github.com/open-orchestra/open-orchestra-display-bundle/pull/158
https://github.com/open-orchestra/open-orchestra-front-bundle/pull/107
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/938

itkg-alavieille:
:+1:

ApiBundle/Transformer/NodeTransformer.php
```diff
@@ -170,13 +173,12 @@ protected function getPreviewLink($scheme, $alias, $encryptedId)
...
+            'token' => $encryptedId,
+            'aliasId' => $aliasId,
+            'nodeId' => $nodeId
+        );
 
-        if ($alias->getPrefix() != "") {
-            $previewLink['name'] .= '/' . $alias->getPrefix();
-            $routeName = 'open_orchestra_base_node_preview_with_prefix';
-            $parameters['languagePrefix'] = $alias->getPrefix();
-        }
```
itkg-ngilain: I am not for this approach of the problem : by removing the language prefix, you're serving the preview on another site that the one required by the contributor. Even more, maybe on a non existent one if all sites are configured with a prefix.
Moreover, we can expect a lot of bugs in the rendering of blocks depending on the url.

nicolasThal: I do not see the purpose of the language prefix as it is never used.

No block should depend on the url to be displayed. The url is a combination of the site, node tree, and node configuration. The block is not aware of all of those parameters, so it would be really wrong to use the url to modify the display of a block

PR #940, opened 2015-07-13
fix current language switcher

itkg-alavieille:


nicolasThal:
On which case is the current locale not set?

itkg-alavieille:
i don't understand your question , if the current locale not set i use defaultLocale of the context

nicolasThal:
My question was : Where is the current locale set, and what is the use case where you can have no current locale set?

itkg-alavieille:
the current locale is set in adminController (adminAction) if this action hasn't language in these parameters then the currentLocale isn't set.

nicolasThal:
:+1:

PR #945, opened 2015-07-15
fix multiple submit on error in HTML5 validation

itkg-nanne:


itkg-ngilain:
:+1: if it solves the problem

PR #948, opened 2015-07-17
fix value transformer declaration

itkg-rboessel:


nicolasThal:
:-1: I will change the names and move the code at the same time

PR #974, opened 2015-07-29
redirect to datatable after add a content

itkg-nanne:


BackofficeBundle/Controller/ContentController.php
```diff
@@ -107,11 +107,7 @@ public function newAction(Request $request, $contentType)
         if ($this->handleForm($form, $message, $content)) {
             $this->dispatchEvent(ContentEvents::CONTENT_CREATION, new ContentEvent($content));
 
-            return $this->redirect(
-                $this->generateUrl('open_orchestra_backoffice_content_form', array(
-                    'contentId' => $content->getContentId()
-                ))
-            );
+            return $this->render('BraincraftedBootstrapBundle::flash.html.twig');
```
itkg-ngilain: I'm not sure to understand: when a content is created we now see a page with only a flashbag ?

BackofficeBundle/Resources/views/BackOffice/Underscore/flashBag._tpl.twig
```diff
@@ -0,0 +1,3 @@
+<div id="flashbag">
```
nicolasThal: You should give a class instead of an id

itkg-rboessel:
:+1: with class and if test pass after rebase

itkg-rboessel:
:+1:

BackofficeBundle/Resources/views/BackOffice/Underscore/flashBag._tpl.twig
```diff
@@ -0,0 +1,3 @@
+<div class="flashbag">
```
nicolasThal: Isn't there some classes in smartadmin about flashes?

PR #1048, opened 2015-09-10
fix tab view form

itkg-nanne:


BackofficeBundle/Resources/public/coffee/tab/tabViewFormLoader.coffee
```diff
@@ -1,22 +1,8 @@
 tabViewFormLoad = (options) ->
   tabViewClass = appConfigurationView.getConfiguration(options.entityType, 'showTab')
 
+  options. panels = getPanelsLink(options.element.get('links'))
```
itkg-ngilain: maybe it will work better without the space :p

PR #1050, opened 2015-09-14
Dynamic load tree

itkg-nanne:


BackofficeBundle/Resources/public/coffee/backboneRouter.coffee
```diff
@@ -21,8 +21,22 @@ OrchestraBORouter = Backbone.Router.extend(
...
+    lis = $('nav li:has(a' + selector + ')')
+    lis.addClass 'active'
+    lis.each ->
+      ul = $('ul', this)
+      link = $(this).children('a').first()
+      if link.data('subtree')
+        $.ajax
+          url: link.data('subtree')
+          type: "GET"
+          async: false
```
juchi: async ajax calls generally do more harm than good, why using it ?

itkg-nanne: because of openMenu

juchi: what do you mean ?

itkg-nanne: line 39, openMenu. How to show something not loaded?

juchi: I don't understand the problem, we always work with callbacks when dealing with ajax, that shouldn't be an issue here.

itkg-nanne: - Some menus use ajax, some others not. openMenu should be duplicate in case of using callback.
- In case of multiple tree ajax calls with callbacks, we could'nt send once openMenu.

For these reasons, async is here safe and make this functionality more flexible.

juchi: * Duplicating a call to a function is actually the reason we use functions, so we don't duplicate multiple lines
* I'm not sure if we're talking about sibling menu items or cascade calls, but it's always possible to trigger an event when all calls are done

I don't exactly know what is safeness in this context but freezing the JS execution doesn't sound good.
The reason of this change was to avoid page freezes because we have some huge nodes in the media library, so the node rendering will be long, so the ajax request itself will take some time. In the end, the issue will remain.

itkg-nanne:
ok for you @juchi ?

juchi:
yes

PR #1071, opened 2015-09-24
add route generation on status change

nicolasThal:


nicolasThal:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1071
https://github.com/open-orchestra/open-orchestra-front-bundle/pull/124
https://github.com/open-orchestra/open-orchestra-model-interface/pull/116
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/433

nicolasThal:
@itkg-ngilain @itkg-alavieille @itkg-nanne I have updated the pr, only the part on the redirection update is missing for now

ApiBundle/Tests/Functional/Controller/AreaControllerTest.php
```diff
@@ -21,6 +21,9 @@ public function testAreaReverseTransform()
         $json = json_decode($this->client->getResponse()->getContent(), true);
         $area = $json['areas'][1];
         $this->assertSame('myMain', $area['area_id']);
+        if (!array_key_exists(0, $area['areas'])) {
+            $this->markTestIncomplete('Datas have been altered');
+        }
```
itkg-ngilain: More than marking the test as incomplete wouldn't it be more logical to stop the tests and display a message to the user saying that he must reload the fixtures ?
I mean if you want to run the tests, you have to run ALL the tests. That second pass is meaningless if some tests aren't runned.

nicolasThal: I would rather suggest to investigate and find where the datas have been modified instead.

In this case, if I run the test once, it is ok, I can also run this test multiple time, but if I run the whole test suite, the test is failing on the second time.

PR #1073, opened 2015-09-25
Add translation Api and use it in TinyMce for custom buttons

itkg-ngilain:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1073
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/38

### DIFF
--- a/BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
+++ b/BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
@@ -1,7 +1,22 @@
 # ACTIVATE TINYMCE
+
 callback_tinymce_init = null
 
+doCallBack = (editor, view) ->
+
+initParameter = () ->
+
+$(document).on('focusin', (e) ->
+  if ($(e.target).closest(".mce-window").length)
+    e.stopImmediatePropagation();
+)
+
 activateTinyMce = (view, textarea) ->
+  $.ajax
+    url: $("#contextual-informations").data("translation-url-pattern").replace("*domain*", "tinymce")
+    success: (response) ->
+      tinymce.util.I18n.add response.lcoale, response.catalog
+
   do (view, textarea) ->
     initParameter()
     tinymce.create 'tinymce.plugins.BBCodeOrchestraPlugin',
@@ -52,15 +67,8 @@ activateTinyMce = (view, textarea) ->
       doCallBack(editor, view)
       return
     return
+
   if textarea.attr('disabled') == 'disabled'
     initTinyMCE($.extend(true, {}, stfalcon_tinymce_config, {theme: {simple: {readonly: 1}}}))
   else
     initTinyMCE()
-
-doCallBack = (editor, view) ->
-initParameter = () ->
-
-$(document).on('focusin', (e) ->
-  if ($(e.target).closest(".mce-window").length)
-    e.stopImmediatePropagation();
-)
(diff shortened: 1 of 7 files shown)


itkg-alavieille:
:+1:

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
```diff
@@ -1,7 +1,22 @@
...
+$(document).on('focusin', (e) ->
+  if ($(e.target).closest(".mce-window").length)
+    e.stopImmediatePropagation();
+)
+
 activateTinyMce = (view, textarea) ->
+  $.ajax
+    url: $("#contextual-informations").data("translation-url-pattern").replace("*domain*", "tinymce")
+    success: (response) ->
+      tinymce.util.I18n.add response.lcoale, response.catalog
```
nicolasThal: respone.locale

BackofficeBundle/Resources/views/layout.html.twig
```diff
@@ -65,6 +65,8 @@
 
     <script data-pace-options='{ "restartOnRequestAfter": true }' src="{{ asset('js/pace.min.js') }} "></script>
 
+    {{ tinymce_init() }}
```
nicolasThal: Why did you move the tinymce activation before all the js load ?

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
```diff
@@ -1,7 +1,22 @@
 # ACTIVATE TINYMCE
+
 callback_tinymce_init = null
 
+doCallBack = (editor, view) ->
```
nicolasThal: What is the usage of these anonymous empty function ?

BackofficeBundle/Resources/views/layout.html.twig
itkg-ngilain: @nicolasThal : In my first attempt I need this as some scripts in our all.js required tinymce to be loaded. I think it's no more the case, but as it changes nothing to be loaded before or after our all.js I suggest to let it before. As some of our future bundles should be related to tinymce, it's more logical to load tinymce before execute plugins.

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
itkg-ngilain: @nicolasThal : I dind't introduce that lambda function, I've just updated it. It's a legacy from a preview work of @itkg-rboessel and @itkg-nanne, so they are the best placed to answer this.
But what I have understood discussing with @itkg-nanne is that this mechanism is introduced to deal with a specific Orchestra combo: only one media modal for several rich textes on the same form. That way, each rich text can locally run that callback, and so the modal correctly knows its parent.

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
itkg-ngilain: @nicolasThal : It's weird that the code works with that anomaly Oo

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
itkg-ngilain: @nicolasThal : I have read your question too fast and did a half-answer. This callback is empty because it is called in this cms-bundle file for a job required in media-admin. So it does nothing in cms-bundle but it is overriden is media-admin.
I'm not fan of the way it works because it allows only one override for all the bundles, but as it was here before my work and as I have no alternative for the moment, I let this like that, but I suggest a refacto when we'll have a better way of doing this.

itkg-ngilain:
@nicolasThal : update done, are you ok with this and my answers ?

BackofficeBundle/Resources/public/coffee/tinyMCE/activateTinyMCE.coffee
nicolasThal: All right :).

This question was really only for my own knowledge, but I will agree that we open a discussion on a refacto on this part.

nicolasThal:
:+1:

PR #1084, opened 2015-10-01
Dynamize dashboard

itkg-ngilain:


ApiBundle/Controller/DashboardController.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @param Request $request
+     *
+     * @Config\Route("", name="open_orchestra_api_dashboard")
+     * @Config\Method({"GET"})
+     *
+     * @return FacadeInterface
+     */
+    public function listWidgetsAction(Request $request)
```
itkg-alavieille: can you add this action in api controller test

BackofficeBundle/DependencyInjection/OpenOrchestraBackofficeExtension.php
```diff
@@ -67,6 +67,7 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('authorize_status_change.yml');
         $loader->load('authorize_edition.yml');
         $loader->load('restore_entity.yml');
+        $loader->load('dashboard.yml');
```
SOLenG: I think you forgot merge the site's configuration with the yaml

itkg-ngilain: totally right ^^'
I work on that

itkg-ngilain: @SOLenG : after a new reflexion it's not a so good idea.
This file only presents a parameter (an array) describing the widgets used on the dashboard (not the widget availbales). So if the integrator wants to use other widgets for his project, he just has to override that array in his project conf. By merging it with our, we'll force our widgets to be present, so merge is not a good thing for that param.

itkg-ngilain:
@itkg-alavieille : PR updated

itkg-alavieille:
:+1:

BackofficeBundle/DependencyInjection/OpenOrchestraBackofficeExtension.php
nicolasThal: In each case, the configuration that the integrator has set is not analyzed, and in each case the dashboard will contain only the few one we set.

I don't know if the merge is the best solution, but I think that a solution inbetween could be found :)

itkg-ngilain: @nicolasThal : If the integrator alters his app/config/parameters.yml, his settings replace the Orchestra default ones. Do you think we should provide another way to override that parameter ?

itkg-ngilain:
@SOLenG , @nicolasThal : update done

nicolasThal:
:+1:

PR #1085, opened 2015-10-02
simplify relations

nicolasThal:


itkg-ngilain:
@I don't agree with that, open-orchestra-base-bundle and open-orchestra-model-interface are trully require :s

PR #1095, opened 2015-10-08
create route for published nodes only

nicolasThal:


BackofficeBundle/Manager/RouteDocumentManager.php
```diff
@@ -42,15 +42,16 @@ public function __construct(
     }
 
     /**
-     * @param NodeInterface $node
+     * @param NodeInterface $givenNode
      *
      * @return array
      */
-    public function createForNode(NodeInterface $node)
+    public function createForNode(NodeInterface $givenNode)
```
itkg-ngilain: why this renaming ? it sounds weird

nicolasThal: I do this renaiming because the node we are going to use to create the routes will not be the same

PR #1102, opened 2015-10-13
Unit test transformers

SOLenG:


ApiBundle/Tests/Transformer/ContentAttributeTransformerTest.php
```diff
@@ -0,0 +1,49 @@
...
+    protected $contentAttributeString;
+    protected $contentAttribute;
+    protected $transformer;
+
+    /**
+     * set Up
+     */
+    public function setUp()
+    {
+        $this->contentAttribute = Phake::mock('OpenOrchestra\ModelInterface\Model\ContentAttributeInterface');
```
nicolasThal: as the contentAttribute is only used in one test, I would rather declare the mock in the test method

ApiBundle/Tests/Transformer/ContentAttributeTransformerTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace OpenOrchestra\ApiBundle\Tests\Transformer;
+
+use OpenOrchestra\ApiBundle\Transformer\ContentAttributeTransformer;
+use Phake;
+
+/**
+ * Class ContentAttributeTransformerTest
+ */
+class ContentAttributeTransformerTest extends \PHPUnit_Framework_TestCase
```
nicolasThal: You should also add a test for the exception thrown

ApiBundle/Tests/Transformer/ContentAttributeTransformerTest.php
```diff
@@ -0,0 +1,49 @@
...
+    protected $transformer;
+
+    /**
+     * set Up
+     */
+    public function setUp()
+    {
+        $this->contentAttribute = Phake::mock('OpenOrchestra\ModelInterface\Model\ContentAttributeInterface');
+        $this->transformer = new ContentAttributeTransformer();
+        $this->contentAttributeString = 'ContentAttributeString';
```
nicolasThal: same for contentAttributeString

ApiBundle/Tests/Transformer/ContentTransformerTest.php
```diff
@@ -31,24 +34,38 @@ class ContentTransformerTest extends \PHPUnit_Framework_TestCase
...
-
         $this->content = Phake::mock('OpenOrchestra\ModelInterface\Model\ContentInterface');
+        Phake::when($this->content)->getId()->thenReturn(1);
         $this->status = Phake::mock('OpenOrchestra\ModelInterface\Model\StatusInterface');
-        $this->statusId = 'StatusId';
-        Phake::when($this->status)->getId(Phake::anyParameters())->thenReturn($this->statusId);
+        $this->contentData = 'contentData';
+        $this->contentDate = '2015/09/01';
+        $this->contentBool = true;
+        Phake::when($this->content)->getContentId(Phake::anyParameters())->thenReturn($this->contentData);
```
nicolasThal: Those methods are only getters, we can assume that they would not receive a parameter, so the Phake::anyParameters() is useless in this case

nicolasThal: you should put those methods in the test only

ApiBundle/Tests/Transformer/StatusTransformerTest.php
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public function setUp()
+    {
+        $this->content = Phake::mock('OpenOrchestra\ModelInterface\Model\ContentInterface');
+        $this->status = Phake::mock('OpenOrchestra\ModelInterface\Model\StatusInterface');
+        $this->statusable = Phake::mock('OpenOrchestra\ModelInterface\Model\StatusableInterface');
+        $this->statusId = 'StatusId';
+        Phake::when($this->status)->getId(Phake::anyParameters())->thenReturn($this->statusId);
+
+        /** Set router */
```
nicolasThal: comment to suppress

ApiBundle/Tests/Transformer/StatusTransformerTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @dataProvider provideTransformData
+     */
+    public function testTransform($published, $initial, $isGranted)
+    {
+        Phake::when($this->authorizeStatusChangeManager)->isGranted(Phake::anyParameters())->thenReturn($isGranted);
+        $document = Phake::mock('OpenOrchestra\ModelInterface\Model\StatusableInterface');
+
+        Phake::when($this->groupContext)->hasGroup(Phake::anyParameters())->thenReturn(false);
```
nicolasThal: You should also add a test when the hasGroup return true

ApiBundle/Tests/Transformer/ContentTransformerTest.php
```diff
@@ -15,44 +16,35 @@ class ContentTransformerTest extends \PHPUnit_Framework_TestCase
...
-    protected $router;
-    protected $status;
 
     /**
      * Set up the test
      */
     public function setUp()
     {
-        $this->eventDispatcher = Phake::mock('Symfony\Component\EventDispatcher\EventDispatcherInterface');
+        $status = Phake::mock('OpenOrchestra\ModelInterface\Model\StatusInterface');//+
```
itkg-alavieille: you can remove characters //* and in other lines

ApiBundle/Tests/Transformer/ContentAttributeTransformerTest.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * test transform
+     */
+    public function testTransform()
+    {
+        $contentAttribute = Phake::mock('OpenOrchestra\ModelInterface\Model\ContentAttributeInterface');
+        $contentAttributeString = 'ContentAttributeString';
+        Phake::when($contentAttribute)->getName()->thenReturn($contentAttributeString);
+        Phake::when($contentAttribute)->getValue()->thenReturn($contentAttributeString);
+        Phake::when($contentAttribute)->getStringValue()->thenReturn($contentAttributeString);
```
itkg-ngilain: You should return 3 different values instead of an unique one. That way you can check the value are correctly setted in the transformer. By returning the same value, the test can't see an inversion between $facade->name, $facade->value or $facade->stringValue;  

By returning the same value for this 3 attributes you can't check that an inversion has been made in the  You should set 3 different values to check that

ApiBundle/Tests/Transformer/ContentTransformerTest.php
```diff
@@ -60,21 +52,44 @@ public function setUp()
...
+        Phake::when($content)->getContentId()->thenReturn($contentData);
+        Phake::when($content)->getContentType()->thenReturn($contentData);
+        Phake::when($content)->getName()->thenReturn($contentData);
+        Phake::when($content)->getVersion()->thenReturn($contentData);
+        Phake::when($content)->getContentTypeVersion()->thenReturn($contentData);
+        Phake::when($content)->getLanguage()->thenReturn($contentData);
+        Phake::when($content)->isDeleted()->thenReturn($contentBool);
+        Phake::when($content)->isLinkedToSite()->thenReturn($contentBool);
+        Phake::when($content)->getCreatedAt()->thenReturn($contentDate);
+        Phake::when($content)->getUpdatedAt()->thenReturn($contentDate);
```
itkg-ngilain: Here too you should use different values instead of a generic one for all getters

nicolasThal:
@itkg-ngilain pr updated

itkg-ngilain:
:+1:

PR #1123, opened 2015-10-19
1.0 prevent route creation

nicolasThal:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1124

PR #1124, opened 2015-10-19
prevent route creation if no published node exists

nicolasThal:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1123


itkg-alavieille:
:+1:

PR #1127, opened 2015-10-21
Add default language

SOLenG:


itkg-ngilain:
I think there's a problem with this one, in my mind the languages should be remap like this:
. Front : en + fr + de
. Back : en + fr

So you shouldn't have a message.de, and you shouldn't have in other messages.yml a language.es key

SOLenG:
actually no further translation will be provided

PR #1135, opened 2015-10-23
fix voter with group without site

itkg-alavieille:


Backoffice/Context/ContextManager.php
```diff
@@ -93,8 +102,12 @@ public function getAvailableSites()
             foreach ($user->getGroups() as $group) {
                 /** @var SiteInterface $site */
                 $site = $group->getSite();
-                if (!$site->isDeleted()) {
-                    $sites[] = $site;
+                if ($site == null ) {
+                    return $this->siteRepository->findByDeleted(false);
```
nicolasThal: What happen if there is no site on the plateform ?

nicolasThal:
:+1:

PR #1140, opened 2015-10-26
Add role node

itkg-alavieille:
https://github.com/open-orchestra/open-orchestra-base-api-bundle/pull/61


BackofficeBundle/Resources/views/BackOffice/Underscore/widgetDuplicate._tpl.twig
```diff
@@ -1,4 +1,6 @@
-<a class="btn btn-info btn-new-version" target="_blank">
-    <i class="glyphicon glyphicon-pencil"></i>
-    <%= text%>
-</a>
+{% if is_granted('ROLE_ACCESS_CREATE_NODE') %}
```
nicolasThal: Is this button used only for the nodes or also for the content ?

itkg-alavieille: It's used by content and node , there is a mistake

itkg-alavieille:
@nicolasThal

nicolasThal:
:+1:

itkg-ngilain:
@itkg-alavieille , @Thiblef : can you add some functionnal tests with a user having no role to check that the security annotations are not removed from the controllers ?

Backoffice/AuthorizeStatusChange/Strategies/NodeStrategy.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function isGranted(StatusableInterface $document, StatusInterface $toStatus)
+    {
+        if ($document instanceof NodeInterface) {
+            if (!$this->authorizationChecker->isGranted(TreeNodesPanelStrategy::ROLE_ACCESS_UPDATE_NODE)) {
+                return false;
+            }
+        }
+
+        return true;
```
itkg-ngilain: @nicolasThal : I think the lack of "support" method cause a problem here : because the strategy is not concerned by the context, it shouldn't be asked for a go/nogo. Because it is forced to return something (yes), it can have a negative impact on the security layer.
What do you think about it ?

BackofficeBundle/Resources/translations/role.en.yml
```diff
@@ -1,6 +1,9 @@
 open_orchestra_role:
     role_access_general_node: Manager the transversal pages
     role_access_tree_node: Manage pages
+    role_access_create_node: Create pages
+    role_access_update_node: Update pages
+    role_access_delete_node: Delete pages
```
itkg-ngilain: In the roles list given by @nbouquet , all roles don't use the plural forms. Here it should be ``Create a page``, and so on.

BackofficeBundle/Resources/translations/role.fr.yml
```diff
@@ -1,6 +1,9 @@
 open_orchestra_role:
     role_access_general_node: Gérer les pages transverses
     role_access_tree_node: Gérer les pages
+    role_access_create_node: Créer des pages
```
itkg-ngilain: Same here

itkg-alavieille:
@itkg-ngilain add tests

Backoffice/AuthorizeStatusChange/Strategies/NodeStrategy.php
nicolasThal: I have no idea about that.

For me the support method was not really usefull because we only want to know if the element can be modified.

itkg-ngilain: I agree without that strategy, but now we can see that the strategy return a non sense random access, so I think we should introduce a support method to avoid asking a strategy not concerned by the request

itkg-ngilain:
ok with the tests

itkg-alavieille:
@itkg-ngilain update translation and use abstract security checker aware transformer

itkg-ngilain:
:+1:

PR #1143, opened 2015-10-27
Split orchestra and smartadmin

nicolasThal:
@juchi is it ok for you ?


juchi:
:+1: for the orchestralibs part

PR #1175, opened 2015-11-02
fixed editing of node when is status published

SOLenG:
https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/BackofficeBundle/Controller/BlockController.php#L41

The option can be set upstream


itkg-alavieille:
:+1:

PR #1183, opened 2015-11-05
add version open orchestra in header

SOLenG:


SOLenG:
https://github.com/open-orchestra/open-orchestra-base-bundle/pull/52

nicolasThal:
As the version is in a const, you could display const in the twig http://twig.sensiolabs.org/doc/tests/constant.html

SOLenG:
@nicolasThal method fix

nicolasThal:
:+1:

PR #1193, opened 2015-11-09
limit ajax abort by method type request

SOLenG:


BackofficeBundle/Resources/public/coffee/setUpCallAjax.coffee
```diff
@@ -8,12 +8,14 @@ $(document).ready ->
         redirectUrl = appRouter.generateUrl 'showDashboard'
         displayMenu(redirectUrl)
       else
-        xhrFifo.push(xhr)
+        xhrFifo.push({xhr:xhr, method: settings.type.toUpperCase()})
```
nicolasThal: I would not put the `toUpperCase` here. You should only use it for the comparaison

BackofficeBundle/Resources/public/coffee/setUpCallAjax.coffee
```diff
@@ -8,12 +8,14 @@ $(document).ready ->
...
         context = settings.context
         displayLoader(context.button) if context != undefined && context.button != undefined
     abortXhr: ->
       for i of xhrFifo
-        xhrFifo[i].abort()
+        if xhrFifo[i].method != 'POST'
```
nicolasThal: If my memory is good, we said that the only type or cancellable requests where the 'GET' requests

juchi:
ping @nicolasThal 
As discussed, this seems like a not-so-good idea. I don't think you should choose which kind of request you abort. I actually think you should not abort any ajax request (at least application-wide). This needs to be further discussed on your side but the abortXhr() method is in my opinion not suited for use in any application.

itkg-ngilain:
@nicolasThal , @juchi : to my mind, some requests are abortable, some other not. With this point of view, developpers should tag abortable requests and only that requests should be stocked.

juchi:
For me it doesn't really solve the problem, because a request than can be aborted should only be aborted (or ignored) by the component that triggered it.

itkg-ngilain:
@SOLenG : is this PR still valid ?

SOLenG:
@itkg-ngilain not realy. I go closing this pr.

PR #1194, opened 2015-11-09
1.0-add abort xhr in data table search

itkg-alavieille:


nicolasThal:
:+1:

BackofficeBundle/Resources/public/coffee/table/TableviewCollectionView.coffee
```diff
@@ -221,6 +222,7 @@ TableviewCollectionView = OrchestraView.extend(
         else if $.isPlainObject(conf.data)
           $.extend(request, conf.data)
 
+        $.ajaxSetup().abortXhr()
```
juchi: It seems that all currently running xhr will be stopped by this line. I'm not sure this is safe. This could end up in a bug if another component on the page is running it's own request.

It would be safer to stop only the grid search requests, or even better, to ignore the results from the outdated requests.
The second approach is IMHO better because it still allows a previous request to get through and to give the look'n'feel of a realtime search. If you abort all requests you cannot see a transitional state during the search.
Also, there is no gain on the server side to abort the request so that wouldn't cost performance.

itkg-alavieille: I create a revert https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1196
We'll talk at the next ceremony

PR #1200, opened 2015-11-09
add a prevent calls form page if click on +/- symbol navigation

SOLenG:


nicolasThal:
I strongly disagree with this modification.

PR #1211, opened 2015-11-13
prevent url load on collapse sign

SOLenG:


BackofficeBundle/Resources/public/smartadmin/js/app.js
```diff
@@ -762,6 +762,11 @@ $.fn.extend({
 				} // end else
 			} // end if
 		});
+
+		// prevent url load on extends accordion menu by click on collapse sign
```
itkg-canne: This file should not be modified as it is part of an external component. Is it possible to put this code in an open-orchestra script file ?

itkg-ngilain: I totally agree, the directory public/smartadmin will be override on the next version of smartadmin, and your code will be then lost.

BackofficeBundle/Resources/public/smartadmin/js/app.js
```diff
@@ -762,6 +762,11 @@ $.fn.extend({
 				} // end else
 			} // end if
 		});
+
+		// prevent url load on extends accordion menu by click on collapse sign
+		$this.find(".collapse-sign").on("click", function(event) {
```
itkg-ngilain: I suggest to use a selector argument to restrict the listening to the nav element

itkg-canne:
I take a look on this problem. I think that the preventing of default action should be done by the component that instantiate the jarvismenu. I think that it can be possible that, in some case, we want to both open the item and go to the linked anchor.
So I think we should check, directly in our custom click listener, if the event target was on the +/- element or directly on the 'a' element, and then choose what to do.

itkg-ngilain:
:+1: but I think you have a problem with your rebase

PR #1259, opened 2015-12-01
Fix RoleCollector for node roles

Thiblef:


Backoffice/Collector/RoleCollector.php
```diff
@@ -77,9 +77,15 @@ public function getRolesByType($type)
     {
         $selectedRoles = array();
 
-        foreach ($this->roles as $role =>  $translation) {
+        foreach ($this->roles as $role => $translation) {
             if (preg_match('/role_access_([^_]*_)?' . $type . '/i', $role)) {
-                $selectedRoles[$role] = $translation;
+                if ($type == 'node') {
```
nicolasThal: yoda notation

itkg-alavieille:
i'm not agree with this refacto, i think than it's not the role collector who should realize this check, why not rename the different role of nodes more specifically.

Thiblef:
@itkg-alavieille  I agree. We can rename the 'role_access_general_node' with 'role_access_tree_general_node' like the node role 'role_access_tree_node'. 
And add tests in the RoleCollectorTest to avoid other problems with role names.

Thiblef:
@itkg-alavieille @nicolasThal

itkg-nanne:
:+1:

Backoffice/NavigationPanel/Strategies/GeneralNodesPanelStrategy.php
```diff
@@ -11,7 +11,7 @@
  */
 class GeneralNodesPanelStrategy extends AbstractNavigationPanelStrategy
 {
-    const ROLE_ACCESS_GENERAL_NODE = 'ROLE_ACCESS_GENERAL_NODE';
```
nicolasThal: I think that you should keep and deprecate this constant.

It might be used somewhere

Thiblef:
@nicolasThal @itkg-alavieille @itkg-nanne

Backoffice/Collector/RoleCollector.php
```diff
@@ -77,8 +77,8 @@ public function getRolesByType($type)
     {
         $selectedRoles = array();
 
-        foreach ($this->roles as $role =>  $translation) {
-            if (preg_match('/role_access_([^_]*_)?' . $type . '/i', $role)) {
+        foreach ($this->roles as $role => $translation) {
+            if (preg_match('/role_access_([^_]*_)?' . $type . '$/i', $role)) {
```
nicolasThal: you could also add the `^` character at the begining of the regex, so you will be sure that nothing more is matched

nicolasThal:
:+1: with the small modification

itkg-ngilain:
:+1:

PR #1276, opened 2015-12-04
fix tab title on local storage refresh

itkg-nanne:


Thiblef:
:+1:

BackofficeBundle/Resources/public/coffee/tab/TabElementFormView.coffee
```diff
@@ -20,4 +20,7 @@ TabElementFormView = OrchestraView.extend(
     @options.domContainer.append @renderTemplate('OpenOrchestraBackofficeBundle:BackOffice:Underscore/backToList',
         listUrl : @options.listUrl
     )
+    
+  onViewReady: ->
```
itkg-alavieille: TabElementFormView is specific at tab form so this update don't work with other tab like those of medias. 

I think there is two solutions : 

 * first create an view (abstract) which add this method and the tabs view extend this view.

 * Second the tabview listen event viewReady of tabs view.

itkg-alavieille:
:+1:

PR #1283, opened 2015-12-08
fix move nodes in navigation tree

aloupfor:


BackofficeBundle/Resources/public/coffee/page/orderNode.coffee
```diff
@@ -1,28 +1,32 @@
...
+    for node in nodes
+      nodeData.push({'node_id': $(node).data('element-id')})
+    nodeCollection = {}
+    nodeCollection['nodes'] = nodeData
+    $.ajax
+      url: ul.data('update-order')
+      method: 'POST'
+      data: JSON.stringify(nodeCollection)
+
+activateOrderNode()
```
nicolasThal: I would rather wait for jQuery to launch the activateOrderNode, what do you think @juchi ?

juchi: I don't know the context, but indeed it generally makes sense to wait for an event to run some code ;)

aloupfor:
@nicolasThal

nicolasThal:
:+1:

itkg-ngilain:
I'm ok with your refacto except for one point: are you sure that the functionnality is ok in the case of a F5 ? I think the ``displayMenu`` function is only called on demand (mainly to refresh the menu). I think it is not called when displaying the page for the first time.

aloupfor:
@itkg-ngilain 
Indeed the `displayMenu` is called when the menu is reloaded. That is why I add the following lines at the end of the `orderNode.coffee` file.

```coffee
jQuery ->
  activateOrderNode()
  return
```
This keeps the previous behavior which was that the nodes could be sorted in a new order after a simple page load.

itkg-ngilain:
Oops, it seems I've read your PR too fast ^^ That looks fine to me so :)
:+1:

PR #1295, opened 2015-12-11
refacto datatable

itkg-alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1295
https://github.com/open-orchestra/open-orchestra/pull/752


BackofficeBundle/Resources/public/coffee/viewConfigurator.coffee
```diff
@@ -11,7 +11,7 @@ OrchestraViewConfigurator = ->
     'addRestoreButtonAction': TableviewRestoreAction
     'addConfigurationButton': PageConfigurationButtonView
     'showTableCollection': TableviewCollectionView
-    'showTableHeader': TableviewSearchHeader
+    #'showTableHeader': TableviewSearchHeader
```
aloupfor: should this line stay as a comment?

itkg-alavieille:
@aloupfor pr updated

BackofficeBundle/Resources/public/coffee/dataTable/DataTableView.coffee
```diff
@@ -0,0 +1,298 @@
+OpenOrchestraDataTableChannel = new (Backbone.Wreqr.EventAggregator)
```
nicolasThal: This channel could have been added to the OpenOrchestra js namespace (see groupTreeLib.coffee)

BackofficeBundle/Resources/public/coffee/dataTable/header/searchField/BooleanFieldSearchView.coffee
```diff
@@ -0,0 +1,41 @@
...
+    )
+    $('.search-column', @$el).val(@options.api.column(@options.columnIndex).search())
+    @insertFieldInHeader()
+
+    return @
+
+((tableFieldViewConfigurator) ->
+  tableFieldViewConfigurator.boolean = BooleanFieldSearchView
+  return
+) window.tableFieldViewConfigurator = window.tableFieldViewConfigurator or {}
```
nicolasThal: the `tableFieldViewConfigurator` could be in the `OpenOrchestra` namespace

GruntTasks/Options/concat.orchestrajs.js
```diff
@@ -89,15 +89,19 @@ module.exports = {
         'web/built/openorchestrabackoffice/js/tab/TabElementFormView.js',
         'web/built/openorchestrabackoffice/js/tab/tabViewFormLoader.js',
 
+        //--[ DATATABLE ]--//
+        'web/built/openorchestrabackoffice/js/dataTable/plugins/InputFullPagination.js',
```
nicolasThal: Is it possible that all the files in the dataTable folder can be loaded without regarding their order ?

BackofficeBundle/Resources/public/coffee/dataTable/DataTableView.coffee
```diff
@@ -0,0 +1,298 @@
...
+    return @
+
+  ###*
+  * @param {object} e jquery event
+  * @param {Object} settings DataTable settings
+  * @param {boolean} processing
+  ###
+  processingData : (e, settings, processing) ->
+    message = settings.oLanguage.sProcessing
+    domProcessing = $("<div class='dataTables_processing'><i class='fa fa-cog fa-spin'></i> "+message+"</div>")
```
itkg-ngilain: Is there any solution to remove this html from the js ? Using an underscore template for instance

GruntTasks/Options/concat.orchestrajs.js
itkg-alavieille: Yes, except for the InputFullPagination because it's a plugin of dataTable

nicolasThal: all right, so you could load all the files in the header folder in one line ?

itkg-alavieille:
@nicolasThal @itkg-ngilain

nicolasThal:
:+1:

BackofficeBundle/Resources/public/coffee/dataTable/DataTableChannel.coffee
```diff
@@ -0,0 +1,11 @@
...
+* @channel DataTable
+* Event available
+*  - clearCache(tableId) clearCache of DataTable
+*  - draw(tableId)
+###
+(($, OpenOrchestra, DataTable) ->
+  DataTable.Channel = new (Backbone.Wreqr.EventAggregator)
+) jQuery,
+  window.OpenOrchestra = window.OpenOrchestra or {} ,
+  window.OpenOrchestra.DataTable = window.OpenOrchestra.DataTable or {}
```
juchi: Why are you providing the main and sub namespaces in the function ? You could go with OpenOrchestra and create DataTable inside the function if it doesn't exist, it cleaner and you'd avoid some logic in the main function call.

BackofficeBundle/Resources/public/coffee/dataTable/header/searchField/BooleanFieldSearchView.coffee
```diff
@@ -0,0 +1,44 @@
...
+
+    return @
+
+(($, OpenOrchestra, DataTable, ViewFieldConfigurator) ->
+  ViewFieldConfigurator.boolean = BooleanFieldSearchView
+  return
+) jQuery,
+  window.OpenOrchestra = window.OpenOrchestra or {} ,
+  window.OpenOrchestra.DataTable = window.OpenOrchestra.DataTable or {} ,
+  window.OpenOrchestra.DataTable.ViewFieldConfigurator = window.OpenOrchestra.DataTable.ViewFieldConfigurator or {}
```
juchi: same remark here

juchi:
Considering my remarks about namespaces and the amount of class and namespaces you're going to have, it may be a good time to start thinking about the global javascript architecture. The new classes being created are still in the global window object (unless i missed some other change), maybe they should be namespaced and/or go inside a limited scope. That would imply to modify the build process too.

itkg-alavieille:
@juchi, yes for the namespace we have a ticket to do a refacto on javascript files with the use the namespaces especially

itkg-ngilain:
@itkg-alavieille: your jsdoc seems to be formatted in several different ways

itkg-ngilain:
:+1: when comment updated

juchi:
You have a pattern that appears several times here, looking more or less like this :
```
class myClass
  // class declaration goes here
(($, OpenOrchestra) ->
  OpenOrchestra.namespace = {} if not OpenOrchestra.namespace?
  // maybe some subnamespaces
  OpenOrchestra.namespace.classAlias = myClass
) jQuery,
  window.OpenOrchestra = window.OpenOrchestra or {}
```

It's a bit strange because an anonymous function is used only to declare subnamespaces, which you could do in the global space with the exact same result (from the userland perception).
The issue is that there is no encapsulation of the class (myClass in my example), it's still accessible from anywhere in the app because it has global scope (and also through your namespaced alias, but the user is not forced to use it)
The reason we use anonymous functions for declarations is that the variables are not accessible from outside so they get private, and then we can give access to some of them through namespaces (which is what is done in my illustration when creating the "classAlias" key).
Long things short : if you want the user to access your class only through the namespace, the pattern should look like (with JS code):

```
(function (OpenOrchestra) {
class myClass {}
OpenOrchestra.myNamespace.myAlias = myClass;
}(window.OpenOrchestra = window.OpenOrchestra || {} ));
```

itkg-alavieille:
@juchi like that ?

juchi:
yes, this make more sense :+1:

PR #1301, opened 2015-12-15
Open nav after content type

itkg-nanne:


BackofficeBundle/Resources/public/coffee/ContentTypeFormView.coffee
```diff
@@ -1,8 +1,9 @@
 ContentTypeFormView = FullPageFormView.extend(
 
   onViewReady: ->
-    displayMenu() if @options.submitted
-
+    if @options.submitted
+      displayRoute = appRouter.generateUrl('listEntities', entityType: @options.entityType)
+      displayMenu(undefined, displayRoute)
```
itkg-ngilain: @itkg-nanne : Why don't you simply pass the displayRoute as the first argument ?

itkg-nanne: to disable Backbone.history.navigate at the end of displayMenu() @itkg-ngilain

itkg-ngilain: Ok, I see better now. I find weird to call a function with a route param sometimes as first argument, and sometimes as second, but in this case with no first parameter. Maybe you could change the function signature into something like ``displayMenu = (route, disableNavigate)``, with ``disableNavigate`` being a boolean.
That way the route is always the first parameter, and if required you can force to disable the ``Backbone.history.navigate``

itkg-nanne:
ok for you @itkg-ngilain ?

SOLenG:
:+1:

BackofficeBundle/Resources/public/coffee/orchestraLib.coffee
```diff
@@ -29,12 +29,9 @@ opts =
...
 
-displayMenu = (route) ->
-  selectedPath = undefined
-  if typeof route isnt "undefined"
-    selectedPath = "#" + route
-  else
-    selectedPath = "#" + Backbone.history.fragment
+displayMenu = (route, refresh) ->
+  selectedPath = "#" + (route || Backbone.history.fragment)
+  refresh = refresh || (typeof route == "undefined")
```
itkg-alavieille: you can use coffee typo : `` refresh = refresh ||  route?

itkg-nanne: will verify if route not null, won't work with undefined. tested @itkg-alavieille

itkg-ngilain:
:+1:

PR #1302, opened 2015-12-15
Add date picker

itkg-nanne:


nicolasThal:
:+1:

BackofficeBundle/Resources/public/coffee/orchestraLib.coffee
```diff
@@ -165,6 +165,7 @@ activateHidden = (hidden) ->
 
 #ACTIVATE FORM JS
 activateForm = (view, form) ->
+  runAllForms()
```
itkg-alavieille: there are not conflict with other activation like select2, because in runAllForms they are an activation of select2 which is diffrent from our ?

itkg-alavieille: why not call only function datepicker ?

PR #1306, opened 2015-12-16
add error message if the form is not valid

nicolasThal:


Thiblef:
:+1:

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
```diff
@@ -1,5 +1,18 @@
 {% extends 'BraincraftedBootstrapBundle:Form:bootstrap.html.twig' %}
 
+{% block form %}
```
itkg-alavieille: why not use the same mecanic than success message, with flash bag ?

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
```diff
@@ -1,5 +1,18 @@
...
+    {% spaceless %}
+        {% if false == form.vars.valid %}
+            <div class="alert alert-danger">
+                There are errors in the form
```
itkg-ngilain: @nicolasThal : You should translate the message

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
itkg-ngilain: I'm not a big fan of the flashbag in this context. But can't we use ``{{ form_error(form) }}``

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
nicolasThal: I didn't took the time to do it :). I was more interested with the comment on how I did this :)

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
nicolasThal: I didn't want to use a flashbag in this case because it would have implied a modification of the php code. By using this method, I ensure that every time a form has an error, there will be an error message.

@itkg-ngilain I will check the `form_errors` block

nicolasThal: @itkg-ngilain in the case of the datetime error, there are no global errors, this way I cannot use the block `form_errors`

nicolasThal:
@itkg-ngilain @itkg-alavieille pr updated

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
itkg-alavieille: in the handleForm when form is not valid if you add a global error message in form, it will go in ``form_error`` no ?

nicolasThal: no, in the `form_start` block, it will only add a class `has-global-errors` to the form. But this class is added only if they are form related errors. In my case, there is only a field error.

itkg-alavieille:
:+1:

PR #1312, opened 2015-12-18
Update inherit node group role

itkg-alavieille:
https://github.com/open-orchestra/open-orchestra-base-api-bundle/pull/71

### DIFF
--- a/ApiBundle/Tests/Transformer/NodeGroupRoleTransformerTest.php
+++ b/ApiBundle/Tests/Transformer/NodeGroupRoleTransformerTest.php
@@ -20,6 +20,8 @@ class NodeGroupRoleTransformerTest extends \PHPUnit_Framework_TestCase
     protected $context;
     protected $roleCollector;
     protected $nodeGroupRoleClass;
+    protected $nodeRepository;
+    protected $currentSiteManager;
 
     /**
      * Set up the test
@@ -30,10 +32,17 @@ class NodeGroupRoleTransformerTest extends \PHPUnit_Framework_TestCase
         Phake::when($this->roleCollector)->hasRole(Phake::anyParameters())->thenReturn(true);
 
         $this->nodeGroupRoleClass = 'OpenOrchestra\GroupBundle\Document\NodeGroupRole';
-
+        $this->nodeRepository = Phake::mock('OpenOrchestra\ModelInterface\Repository\NodeRepositoryInterface');
         $this->context = Phake::mock('OpenOrchestra\BaseApi\Transformer\TransformerManager');
+        $this->currentSiteManager = Phake::mock('OpenOrchestra\BaseBundle\Context\CurrentSiteIdInterface');
 
-        $this->transformer = new NodeGroupRoleTransformer($this->facadeClass, $this->nodeGroupRoleClass, $this->roleCollector);
+        $this->transformer = new NodeGroupRoleTransformer(
+            $this->facadeClass,
+            $this->nodeGroupRoleClass,
+            $this->roleCollector,
+            $this->nodeRepository,
+            $this->currentSiteManager
+        );
         $this->transformer->setContext($this->context);
     }
 
@@ -82,8 +91,8 @@ class NodeGroupRoleTransformerTest extends \PHPUnit_Framework_TestCase
     public function provideTransformData()
     {
         return array(
-            array('foo', 'bar', NodeGroupRoleInterface::ACCESS_GRANTED),
-            array('bar', 'foo', NodeGroupRoleInterface::ACCESS_DENIED),
+            array('foo', 'bar', NodeGroupRoleInterface::ACCESS_GRANTED, true),
+            array('bar', 'foo', NodeGroupRoleInterface::ACCESS_DENIED, false),
         );
     }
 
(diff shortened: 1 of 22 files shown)


ApiBundle/Transformer/NodeGroupRoleTransformer.php
```diff
@@ -52,25 +66,43 @@ public function transform($nodeGroupRole)
...
+     * @param GroupInterface|null $group
      *
      * @throws RoleNotFoundHttpException
-     * @return mixed
+     * @throws TransformerParameterTypeException
+     *
+     * @return null|NodeGroupRoleInterface
      */
-    public function reverseTransform(FacadeInterface $facade, $source = null)
+    public function reverseTransform(FacadeInterface $nodeRoleFacade, $group = null)
```
nicolasThal: I would modify the signature by adding a parameter : ($nodeRoleFacade, $source = null, $group = null).

This way you will modify the source which is given by the caller of the method and you won't have to find it inside this method

itkg-alavieille: but the method of signature of TransformerInterface take only two parameters

nicolasThal: yes, but you can add one if it is still comptatible ( like $group = null at the end)

itkg-ngilain: I think this will introduce a larger problem. If an integrator create it's own ``NodeGroupRoleTransformer``, he may create it with only 2 parameters. But the ``GroupTransformer`` will still call the ``NodeGroupRoleTransformer->reverseTransform`` with 3 parameters, leading to an exception.

GroupBundle/EventListener/UpdateNodeGroupRoleListener.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function preUpdate(PreUpdateEventArgs $event)
+    {
+        $document = $event->getDocument();
+        $uow = $event->getDocumentManager()->getUnitOfWork();
+        if (
+            $document instanceof NodeGroupRoleInterface &&
+            $event->hasChangedField("accessType")
+        ) {
+            $parentAssociation = $uow->getParentAssociation($document);
```
itkg-ngilain: @itkg-alavieille, @nicolasThal : As this listener seems linked with doctrine ODM, maybe we should find a way to explicitally show that. Maybe by renaming the listener ?

GroupBundle/Tests/EventSubscriber/UpdateNodeGroupRoleMoveNodeSubscriberTest.php
```diff
@@ -0,0 +1,226 @@
...
+     * @var UpdateNodeGroupRoleMoveNodeSubscriber
+     */
+    protected $subscriber;
+
+    /**
+     * Set up the test
+     */
+    public function setUp()
+    {
+
```
itkg-ngilain: empty line

ApiBundle/Transformer/NodeGroupRoleTransformer.php
nicolasThal: https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/ApiBundle/Transformer/AreaTransformer.php#L223

Regarding what is done in others transformers (like the node one).e

I would prefer to let three parameters.

If an integrator will override it, i will have to respect the signature given by the class overrided

GroupBundle/EventListener/UpdateNodeGroupRoleListener.php
nicolasThal: Yes maybe.

As a matter of fact, I think that it is not so great to rely our code on doctrine listeners because they are specific to doctrine odm's events.

@itkg-alavieille would it be possible to make the same modification in a listener listening on nodeGroupRole events ?

itkg-alavieille: this listener use method getParentAssociation of unit of work so doctrine odm is required

nicolasThal: all right so, the listener is needed

ApiBundle/Transformer/NodeGroupRoleTransformer.php
itkg-alavieille: i disagree with this, what is the interest of the interface if you don't respect the signatures?

itkg-ngilain: I agree with @itkg-alavieille : an integrator developping a new class must respect the interface, our presenting only 2 params. If we should look at the class, the interface is useless or in our case non operationnal.

Plus there's a great problem with https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/ApiBundle/Transformer/AreaTransformer.php#L223 : if the $node is trully null, it severly crashes. I'd be for a refacto of this part.
A basic solution should be to put a third mixed param to allow each class to pass an optional parameter.

ApiBundle/Tests/Transformer/NodeGroupRoleTransformerTest.php
```diff
@@ -82,8 +91,8 @@ public function testTransform($nodeId, $role, $accessType)
     public function provideTransformData()
     {
         return array(
-            array('foo', 'bar', NodeGroupRoleInterface::ACCESS_GRANTED),
-            array('bar', 'foo', NodeGroupRoleInterface::ACCESS_DENIED),
+            array('foo', 'bar', NodeGroupRoleInterface::ACCESS_GRANTED, true),
+            array('bar', 'foo', NodeGroupRoleInterface::ACCESS_DENIED, false),
```
SOLenG: among the methods using this provider, testTransform ( https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/update_inherit_node_group_role/ApiBundle/Tests/Transformer/NodeGroupRoleTransformerTest.php#L73 ) accepts only 3 arguments

ApiBundle/Transformer/NodeGroupRoleTransformer.php
itkg-alavieille: @nicolasThal

nicolasThal: Todo :

 - Create a `TransformerWithContentInterface` with the new method
 - Add the interface to the `NodeGroupRoleTransformer`
 - Check in the `GroupTransformer` if the asked transformer implements the interface

itkg-alavieille:
@nicolasThal @SOLenG @itkg-ngilain

ApiBundle/Transformer/GroupTransformer.php
```diff
@@ -106,14 +108,21 @@ public function transform($group)
      * @param GroupInterface|null $group
      *
      * @return mixed
+     * @throws
      */
     public function reverseTransform(FacadeInterface $facade, $group = null)
     {
         foreach ($facade->getNodeRoles() as $nodeRoleFacade) {
-            $group->addNodeRole($this->getTransformer('node_group_role')->reverseTransform(
+            $transformer = $this->getTransformer('node_group_role');
```
nicolasThal: You should do the test outsite the `foreach` loop

ApiBundle/Transformer/NodeGroupRoleTransformer.php
```diff
@@ -5,28 +5,43 @@
 use OpenOrchestra\ApiBundle\Exceptions\HttpException\RoleNotFoundHttpException;
 use OpenOrchestra\ApiBundle\Exceptions\TransformerParameterTypeException;
 use OpenOrchestra\Backoffice\Collector\RoleCollectorInterface;
+use OpenOrchestra\BackofficeBundle\Model\GroupInterface;
 use OpenOrchestra\BackofficeBundle\Model\NodeGroupRoleInterface;
 use OpenOrchestra\BaseApi\Facade\FacadeInterface;
 use OpenOrchestra\BaseApi\Transformer\AbstractTransformer;
+use OpenOrchestra\BaseApi\Transformer\TransformerWithContextInterface;
```
nicolasThal: I would rather keep this interface in the ApiBundle because it is only related to this case.

Moreover, I would name it `TransformerWithGroupContextInterface` instead

itkg-alavieille:
@nicolasThal

ApiBundle/Transformer/NodeGroupRoleTransformer.php
```diff
@@ -52,25 +66,43 @@ public function transform($nodeGroupRole)
...
      * @throws RoleNotFoundHttpException
-     * @return mixed
+     * @throws TransformerParameterTypeException
+     *
+     * @return null|NodeGroupRoleInterface
      */
-    public function reverseTransform(FacadeInterface $facade, $source = null)
+    public function reverseTransformWithContext(GroupInterface $group, FacadeInterface $nodeRoleFacade, $source = null)
     {
+        if (!$group instanceof GroupInterface) {
```
nicolasThal: This check is useless with the proper method  signature

ApiBundle/Transformer/NodeGroupRoleTransformer.php
```diff
@@ -52,25 +66,43 @@ public function transform($nodeGroupRole)
...
      * @param NodeGroupRoleInterface|null $source
      *
      * @throws RoleNotFoundHttpException
-     * @return mixed
+     * @throws TransformerParameterTypeException
+     *
+     * @return null|NodeGroupRoleInterface
      */
-    public function reverseTransform(FacadeInterface $facade, $source = null)
+    public function reverseTransformWithContext(GroupInterface $group, FacadeInterface $nodeRoleFacade, $source = null)
```
nicolasThal: I would call the method `reverseTransformWithGroup` to show that you have to give a group and this way you could differentiate all the following method names

ApiBundle/Transformer/TransformerWithGroupContextInterface.php
```diff
@@ -0,0 +1,22 @@
...
+namespace OpenOrchestra\ApiBundle\Transformer;
+
+use OpenOrchestra\BackofficeBundle\Model\GroupInterface;
+use OpenOrchestra\BaseApi\Facade\FacadeInterface;
+use OpenOrchestra\BaseApi\Transformer\TransformerInterface;
+
+/**
+ * Interface TransformerWithGroupContextInterface
+ */
+interface TransformerWithGroupContextInterface extends TransformerInterface
```
nicolasThal: I would also change the part `GroupContext` as it is already used for a different part of the application

itkg-alavieille:
@nicolasThal

ApiBundle/Transformer/GroupTransformer.php
```diff
@@ -106,14 +107,21 @@ public function transform($group)
...
      *
      * @return mixed
+     * @throws
      */
     public function reverseTransform(FacadeInterface $facade, $group = null)
     {
+        $transformer = $this->getTransformer('node_group_role');
         foreach ($facade->getNodeRoles() as $nodeRoleFacade) {
-            $group->addNodeRole($this->getTransformer('node_group_role')->reverseTransform(
+            if (!$transformer instanceof TransformerWithGroupInterface) {
```
nicolasThal: you should do the check on the transformer outsite of the foreach

GroupBundle/EventSubscriber/UpdateNodeGroupRoleMoveNodeSubscriber.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * @param NodeInterface $node
+     * @param array         $nodeRole
+     * @param array         $groups
+     *
+     * @throws NodeGroupRoleNotFoundException
+     */
+    protected function updateNodeGroupRoleTree(NodeInterface $node, array $nodeRole, array $groups) {
```
nicolasThal: psr with the `{`

nicolasThal:
:+1: with the checkstyle

PR #1317, opened 2015-12-18
fix display long title on nav, header, breadcrumb

SOLenG:


BackofficeBundle/Resources/public/coffee/orchestraLib.coffee
```diff
@@ -11,7 +11,7 @@ renderPageTitle = ->
 
 # ADD CUSTOM JARVIS WIDGET
 addCustomJarvisWidget = (widget) ->
-  $(widget).insertAfter($(".js-widget-title"))
+  $(widget).insertBefore($(".js-widget-title"))
```
itkg-ngilain: I think it's a trully big change that goes far further than the +/- problem. Do you have check that the other widgets are not broken ?
I'm very doubtfully about this big change.

BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
```diff
@@ -1,9 +1,27 @@
 .js-widget-title {
     margin-left: 10px;
+    display: block !important;
```
itkg-alavieille: this property isn't necessary, you should use it because in the template the div with class  .js-widget-title, has also the class display inline

BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
```diff
@@ -1,9 +1,27 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    margin-right: 27px;
```
itkg-alavieille: i think than this property isn't necessary

BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
```diff
@@ -1,9 +1,27 @@
 .js-widget-title {
     margin-left: 10px;
+    display: block !important;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
itkg-alavieille: if you add this property, version of node is'nt visible

BackofficeBundle/Resources/public/less/smartadminPatches/nav.less
```diff
@@ -16,5 +16,38 @@ nav {
...
+      white-space: nowrap;
+    }
+    span.menu-item-parent{
+      display: inline-block;
+      overflow: initial;
+    }
+  }
+}
+
+ol.breadcrumb {
```
itkg-alavieille: i think it's better to add this part in other file because this doesn't concern navigation

BackofficeBundle/Resources/public/less/smartadminPatches/nav.less
```diff
@@ -16,5 +16,38 @@ nav {
...
+      position: absolute !important;
+      right: 10px;
+      top: 10px;
+    }
   }
 }
+
+#left-panel nav ul li {
+  > a {
+    b.collapse-sign {
```
itkg-alavieille: this propery is already set by this rule (> li > a b  in line 19)

BackofficeBundle/Resources/public/coffee/orchestraLib.coffee
SOLenG: @itkg-ngilain I agree with you. This was the point that was bothering me the most in my own pr. What decided me was that the semantic result is identical to the header of modals. This change allowed only to maintain the buttons on the top right of the header.

nicolasThal:
:+1:

PR #1328, opened 2015-12-31
fix test for multirun

nicolasThal:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1328
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/502


SOLenG:
:+1:

PR #1341, opened 2016-01-08
Add area transformer interface

SOLenG:


ApiBundle/Transformer/AreaTransformerInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+
+use OpenOrchestra\BaseApi\Facade\FacadeInterface;
+use OpenOrchestra\ModelInterface\Model\AreaInterface;
+use OpenOrchestra\ModelInterface\Model\TemplateInterface;
+
+/**
+ * Interface AreaTransformerInterface
+ */
+interface AreaTransformerInterface extends TransformerInterface
```
nicolasThal: I would rather call this interface `TransformerWithTemplateContext` and it should not extends the `TransformerInterface` see : https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1312

itkg-nanne:
:+1:

PR #1344, opened 2016-01-11
Clean route redirection site update

itkg-nanne:


BackofficeBundle/Controller/SiteController.php
```diff
@@ -27,7 +29,7 @@ class SiteController extends AbstractAdminController
     public function formAction(Request $request, $siteId)
     {
         $site = $this->get('open_orchestra_model.repository.site')->findOneBySiteId($siteId);
-
+        $aliases = $site->getAliases()->toArray();
```
nicolasThal: why do you add this line and do the same call on line 46 ?

itkg-nanne: first $aliases load values in datatable, second one, values from form. By this way, i'm able to know if aliases have change. @nicolasThal

nicolasThal: Sorry I didn't saw that you where using it :)

BackofficeBundle/Controller/SiteController.php
```diff
@@ -41,6 +43,9 @@ public function formAction(Request $request, $siteId)
         $form->handleRequest($request);
         $message =  $this->get('translator')->trans('open_orchestra_backoffice.form.website.success');
         if ($this->handleForm($form, $message)) {
+            if($site->getAliases()->toArray() !== $aliases) {
+                $this->dispatchEvent(SiteAliasEvents::SITEALIAS_UPDATE, new SiteAliasEvent($site));
```
nicolasThal: Why do you dispatch a SITEALIAS_UPDATE with an event which contains only the siteId.

I think that you could have bind the route update on the siteUpdate directly

itkg-nanne: Right. but If I use SITE_EVENT, I just receive the new site whithout knowing if aliases were changed or not. In this case, I've got two solutions, change event to pass old site too or refresh route even if aliases didn't change. That's why, I made this choice. @nicolasThal

itkg-nanne:
done @nicolasThal

nicolasThal:
:+1: with tests running

PR #1346, opened 2016-01-12
remove link new page, node tranverse

itkg-alavieille:


BackofficeBundle/Resources/views/Tree/tree-macros.html.twig
```diff
@@ -12,7 +12,10 @@
             >
                 <span>{{ node.name }}</span>
             </a>
-            {% if 'child' in element|keys and (element.child is not empty or is_granted("ROLE_ACCESS_CREATE_NODE"))%}
+            {% if 'child' in element|keys
+               and (element.child is not empty or is_granted("ROLE_ACCESS_CREATE_NODE"))
+               and node.nodeId != constant('TRANSVERSE_NODE_ID', node)
```
nicolasThal: I think that constant should take the fully qualified namespace of the class and then the constant name

itkg-alavieille: @nicolasThal twig can be test constants from object instances, if use the other syntax with the fully namespaces, when the class name change or namespace it is necessary to updated the twig and not in the first syntax

nicolasThal: sorry, I read it to fast :)

But this syntax is ok for me

nicolasThal:
:+1:

PR #1349, opened 2016-01-12
Remove version node transverse

itkg-alavieille:


nicolasThal:
:+1: but when the test is fixed

PR #1364, opened 2016-01-15
Add concurrency view

itkg-nanne:


BackofficeBundle/Resources/public/coffee/OrchestraView.coffee
```diff
@@ -94,4 +94,16 @@ OrchestraView = Backbone.View.extend(
...
       path: path.duplicate
     ) if element.get('links')._self_duplicate
+
+  addConcurrency: ->
+    if typeof @currentRankView == 'undefined'
+      @oldRender = @render
+      @currentRankView = 0
+    else
+      @currentRankView = @currentRankView + 1
+    @render = () ->
```
itkg-alavieille: why not add an abstract view than the different channel extend it. In this abstract view you could add this this behavior on the render.

itkg-nanne:
done @itkg-alavieille

BackofficeBundle/Resources/public/coffee/widget/AbstractConcurrancyCounter.coffee
```diff
@@ -0,0 +1,17 @@
...
+  addConcurrency: ->
+    if typeof @currentRankView == 'undefined'
+      @oldRender = @render
+      @currentRankView = 0
+    else
+      @currentRankView = @currentRankView + 1
+    @render = () ->
+      prototype = Object.getPrototypeOf(@)
+      if @currentRankView == prototype.currentRankView
+        @oldRender.apply this, arguments
```
SOLenG: I'm a little perturbed, if for a any reason `prototype.currentRankView` has a same value with `@currentRankView` which is defined, so `@oldRender` will not be initialised.

itkg-nanne: That's the point, not doing the render ifnot same currentRank after having define oldRender line 8

itkg-nanne: @oldRender is always defined as @currentRankView is a dynamically defined option and is always undefined as starting state.

SOLenG: if it's the case, so it's :+1: for me

PR #1367, opened 2016-01-18
fix source node with template subarea

itkg-alavieille:


itkg-ngilain:
:+1:

BackofficeBundle/EventSubscriber/NodeTemplateSelectionSubscriber.php
```diff
@@ -47,8 +47,14 @@ public function preSubmit(FormEvent $event)
             ) {
                 $template = $this->templateRepository->findOneByTemplateId($data['nodeTemplateSelection']['templateId']);
                 if (null !== $template) {
-                    $formData->setAreas($template->getAreas());
-                    $formData->setBlocks($template->getBlocks());
+                    foreach($template->getAreas() as $area) {
```
nicolasThal: What will happen if there is more than two levels of area ?

PR #1371, opened 2016-01-18
Remove smartadmin cssconcat from prod environnement

itkg-ngilain:


SOLenG:
it's unfortunable you don't use the "twig loop" to generate a link as it was done before the refacto, else :+1:

itkg-ngilain:
@SOLenG : the refacto is done because that "twig loop" introduces a css bug in a non-debug environnement. In a such environnement, all files are concatenate, and this is what we want to avoid with smartadmin.

nicolasThal:
:+1:

PR #1382, opened 2016-01-26
remove searchable type from field type and inject in subscriber

itkg-nanne:
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/129
[OO-FEATURE] auto-select type of search in datatable for content field type


nicolasThal:
:+1:

PR #1385, opened 2016-01-27
fix NodeGroupRole creation when it already exists

Thiblef:
[OO-BUGFIX] When a node is duplicated there's no more node group role creation


itkg-ngilain:
:+1: but can you alter your PR comment to include the new oo-tags ?

itkg-ngilain:
:+1: but can you alter your PR comment to include the new oo-tags ?

PR #1386, opened 2016-01-27
replace dsiplay menu by refresh menu

itkg-nanne:
[OO-BCBREAK] change way to load dataparameter in refresh menu context
[OO-BUGFIX] disable load dataparemeter in modal context


itkg-ngilain:
:+1:

PR #1388, opened 2016-01-28
fix jarvis widget css

Thiblef:
[OO-BUGFIX] Fix jarvis widget header style in BackOffice

### DIFF
--- a/BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
+++ b/BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
@@ -1,10 +1,10 @@
 .js-widget-title {
     margin-left: 10px;
     display: inline-block;
-    max-width: 90%;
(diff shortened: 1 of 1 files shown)


itkg-ngilain:
:+1:

BackofficeBundle/Resources/public/less/smartadminPatches/widget.less
```diff
@@ -1,10 +1,10 @@
 .js-widget-title {
     margin-left: 10px;
     display: inline-block;
-    max-width: 90%;
```
itkg-alavieille: why remove this line ? this line is used when a the navigator don't support calc property. Like IE9 where calc property isn't totally supported

Thiblef:
http://caniuse.com/#search=calc()
It's not supported only on IE9 when you use it in a background-position. I think this line can be suppressed here.

PR #1396, opened 2016-02-01
add activate form in case of collection prototype

itkg-nanne:
[OO-BUGFIX] Activate tinyMce in collection prototype context
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/135


itkg-alavieille:
:+1:  but think to update MediaFormView also

itkg-ngilain:
:+1:

PR #1433, opened 2016-02-22
fixes #1401 on notification type

### LINKED ISSUE #1401: Support for info / success notification
https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/BackofficeBundle/Resources/public/coffee/orchestraLib.coffee#L247

Notification only supports two type :
* error which are red with an cross icon class
* the others which are yellow with an info icon

The yellow color could be misleading when you use the notification for a feedback. We should at least have a support of info / success for those notifications as it is for flash message.

nicolasThal:
[OO-FEATURE] add notification colors


Thiblef:
You could have add the `info` type in blue like in smartadmin.

nicolasThal:
@Thiblef "info" will be the fallback type

Thiblef:
:+1:

PR #1438, opened 2016-02-24
disable datatable buttons when no access

Thiblef:
[OO-FEATURE] Datatable action buttons are disabled if the user hasn't the rights


itkg-ngilain:
@Thiblef : You should not replace the button type class (btn-warning, btn-danger, etc ...) with the btn-default class. You can disable them with their original style.
Have a look here : http://smartadmin.openorchestra.inte/1.8/DEVELOPER/HTML_Full_Version/buttons.html

Thiblef:
@itkg-ngilain I did it first but I think the information is more visible to the user if we remove color.

itkg-ngilain:
hum ok, as you want.
:+1: then :)

PR #1439, opened 2016-02-24
[OO-FEATURE] Remove max length option on email field type

itkg-ngilain:
[OO-FEATURE] Remove max length option on email field type


Thiblef:
:+1:

itkg-admin:
I propose a default length of 100 : e.g. a majority of businessdecision.com emails has between 30 and 45  letters -> I think that 50 is too short

itkg-ngilain:
@itkg-admin : The required featured was to remove the limit concept on the email field type. Do you want to reintroduce it ?

PR #1440, opened 2016-02-24
Change node order on delete

Thiblef:
[OO-MISC] Deleted nodes order is -1
https://github.com/open-orchestra/open-orchestra-model-interface/pull/164


itkg-ngilain:
:+1:

PR #1449, opened 2016-02-26
Add badge indicator in tab view

clement-michelet:
Works like the title. Just have to add a data attribute `date-badge="foo"` in the first node of the panel to display a badge indicator.

### DIFF
--- a/BackofficeBundle/Resources/public/coffee/tab/TabView.coffee
+++ b/BackofficeBundle/Resources/public/coffee/tab/TabView.coffee
@@ -19,11 +19,11 @@ TabView = OrchestraView.extend(
     $('.js-widget-title', @options.domContainer).html @options.title
 
 
-  addPanel: (title, id, view, active, position) ->
+  addPanel: (title, id, view, active, badge, position) ->
     if @$tabNav? and @$tabContent?
-      @createPanel(title, id, view, active, position)
+      @createPanel(title, id, view, active, badge, position)
     else
-      @options.panels.push {title: title, id: id, view: view, active: active, position: position}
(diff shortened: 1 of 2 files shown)


itkg-ngilain:
We need to talk about this feature before merging it. It may be a little too binded to your project.

We'll study it very quickly. Thank you for your proposition :)

clement-michelet:
This is not related to my project (not requested, not needed), so It's not needed for 1.0.

BackofficeBundle/Resources/public/coffee/tab/TabView.coffee
```diff
@@ -19,11 +19,11 @@ TabView = OrchestraView.extend(
...
 
 
-  addPanel: (title, id, view, active, position) ->
+  addPanel: (title, id, view, active, badge, position) ->
     if @$tabNav? and @$tabContent?
-      @createPanel(title, id, view, active, position)
+      @createPanel(title, id, view, active, badge, position)
     else
-      @options.panels.push {title: title, id: id, view: view, active: active, position: position}
+      @options.panels.push {title: title, id: id, view: view, active: active, badge, position: position}
```
itkg-nanne: You can't use the value badge without a key in associative context.
This key should be use line 57.
As badge is not mandatory, this option should be after position.
It would be easier to use as in https://github.com/open-orchestra/open-orchestra-media-admin-bundle/blob/master/MediaAdminBundle/Resources/public/coffee/SuperboxLoad.coffee Line 29.
@clement-michelet

itkg-ngilain:
@clement-michelet : can you update this pr ? As it seems to be a stalled one, if you don't answer us, we will close it as it's becoming very old.

PR #1452, opened 2016-02-26
Update flage

nicolasThal:


nicolasThal:
https://github.com/open-orchestra/open-orchestra-mongo-libs/pull/20
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/529
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1452
https://github.com/open-orchestra/open-orchestra-model-interface/pull/166

Backoffice/EventSubscriber/UpdateNodeCurrentlyPublishedFlagSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param NodeEvent $event
+     */
+    public function updateFlag(NodeEvent $event)
+    {
+        $node = $event->getNode();
+
+        if ($node->getStatus()->isPublished()) {
```
itkg-ngilain: You could avoid a lot of unnecessaries process if you add a condition here: If the node status is published, but the node is already  currentlyPublished, it's not optimal to do the stuff here, it has always been done previously.

Unless you want to be compatible with older versions of Orchestra, but for this I would privilegiate a doctrine migration script.

Backoffice/EventSubscriber/UpdateNodeCurrentlyPublishedFlagSubscriber.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param NodeInterface $node
+     */
+    protected function updatePublishedFlag(NodeInterface $node)
+    {
+        $publishedNodes = $this->nodeRepository->findCurrentlyPublished($node->getNodeId(), $node->getLanguage(), $node->getSiteId());
+        /** @var NodeInterface $publishedNode */
+        foreach ($publishedNodes as $publishedNode) {
```
itkg-ngilain: I have the same remark as previous: in a normal working platform there can be only one published version per language. And as you get it in the updateFlag method, so I would rather inject it in the updatePublishedFlag method to avoid a second Repo call.

nicolasThal:
@itkg-alavieille pr updated

itkg-nanne:
:+1:

PR #1455, opened 2016-02-29
Remove 'media' from tinymce context menu

itkg-ngilain:
[OO-BUGFIX] Remove native media option in tinymce context menu


nicolasThal:
:+1:

PR #1456, opened 2016-03-03
translate message

djamnazi:


BackofficeBundle/Resources/translations/messages.en.yml
```diff
@@ -94,6 +94,9 @@ open_orchestra_backoffice:
                 date: Date
                 choice: Choice
                 content: Content choice
+                swchoff:
```
itkg-alavieille: buttons are global at all form, so i think it's better to add translation in the same level than form.submit or form.edit etc

itkg-ngilain: I agree with that, you can put them 2 levels higher in the tree

djamnazi:
@itkg-alavieille pr upgated

itkg-alavieille:
:+1:

PR #1459, opened 2016-03-03
hide rootmenu if not submenu

itkg-alavieille:
[OO-BUGFIX] In navigation panel, a root menu is hidden if these submenus aren't visible


itkg-ngilain:
:+1:

PR #1460, opened 2016-03-03
Fix custom field selection on content type creation

itkg-ngilain:
[OO-BUFIX] Fix custom field selection on content type creation


itkg-alavieille:
:+1:

PR #1467, opened 2016-03-08
ad loader in group tab

itkg-nanne:
[OO-FEATURE] add loader in group tab view


BackofficeBundle/Resources/public/coffee/groupTree/FormNodeCollectionView.coffee
```diff
@@ -0,0 +1,15 @@
+###*
```
itkg-alavieille: Same that in open-orchestra-media-admin-bundle

itkg-ngilain:
:+1:

PR #1471, opened 2016-03-09
hide preview button on global pages

djamnazi:
[OO-BUGFIX] hide the preview button on global pages (transverse)

### DIFF
--- a/BackofficeBundle/Resources/views/BackOffice/Underscore/widgetPreviewLink._tpl.twig
+++ b/BackofficeBundle/Resources/views/BackOffice/Underscore/widgetPreviewLink._tpl.twig
@@ -1,4 +1,3 @@
-<% if (_.size(previewLinks)) { %>
 <div class="widget-toolbar">
     <div class="btn-group">
         <button class="btn dropdown-toggle btn-xs txt-color-white bg-color-blue" data-toggle="dropdown">
@@ -16,4 +15,3 @@
         </ul>
     </div>
 </div>
-<% } %>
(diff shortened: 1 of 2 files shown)


BackofficeBundle/Resources/views/BackOffice/Underscore/widgetPreviewLink._tpl.twig
```diff
@@ -1,16 +1,19 @@
+<% if (_.size(previewLinks)) { %>
```
itkg-alavieille: size return a number so can you check if is upper than 0

itkg-alavieille: can you indent the content inside in if

itkg-alavieille:
i think the tag for changelog is [OO-BUGFIX] and not [OO-MISC]

itkg-nanne:
It is not the right way to do this.
If you look at this file : https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/BackofficeBundle/Resources/public/coffee/widget/previewLinkChannel.coffee, a test is done line 2 to know if the preview widget should be shown or not.
This test is no more correct. In global page view.options.node.get('preview_links') is not undefined but gets a 0 size. Certainly a change in the php facade.

djamnazi:
@itkg-alavieille @itkg-alavieille PR updated

itkg-alavieille:
:+1:

PR #1485, opened 2016-03-14
Move template menu entry

itkg-ngilain:
[OO-FEATURE] Move template menu entry from Editorial to Administration


itkg-nanne:
:+1:

PR #1491, opened 2016-03-14
fix searchable contentType

alavieille:
[OO-BUGFIX] Fix searchable option of content type field

https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1491
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/532


itkg-ngilain:
:+1:

PR #1496, opened 2016-03-15
Fix color Picker form type on both context : video block and standard form

itkg-ngilain:
[OO-BUGFIX] Fix color picker form type


alavieille:
:+1:

PR #1498, opened 2016-03-16
Add check role

itkg-nanne:
[OO-BUGFIX] disallow roles with same from and to status


alavieille:
why a subscriber ? and not a validator ?

itkg-nanne:
change done @alavieille

alavieille:
:+1:

PR #1500, opened 2016-03-16
Fix TinyMCE display after form submission

itkg-ngilain:
[OO-BUGFIX] Fix TinyMCE content display after form submission


itkg-nanne:
:+1:

PR #1505, opened 2016-03-17
send only when document is valid

alavieille:
[OO-BUGFIX] The event ``ContentTypeEvents::CONTENT_TYPE_UPDATE`` is dispatched only when content type form is valid


itkg-ngilain:
No OO-TAG ?

:+1:

PR #1513, opened 2016-03-21
fix translation question

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1513
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/176


itkg-ngilain:
:+1:

PR #1521, opened 2016-03-24
refresh menu when a content type is deleted

alavieille:
[OO-FEEATURE] When a content type is navigation menu is refresh


GruntTasks/Targets/concat.orchestra_js.js
```diff
@@ -121,6 +122,9 @@ module.exports = {
     'web/built/openorchestrabackoffice/js/user/UserFormView.js',
     'web/built/openorchestrauseradmin/js/user/*.js',
 
+    //--[ CONTENT TYPE ]--//
+    'web/built/openorchestrabackoffice/js/contentType/TableViewAction.js',
```
itkg-ngilain: I think it would be better to add this to the TableView section, but as you want

itkg-ngilain:
:+1: with or without the little modification

PR #1544, opened 2016-04-04
fix transform area with an other site than current site

alavieille:
[OO-BUGFIX] Fix transform area from another site as the current site


itkg-ngilain:
:+1:

PR #1555, opened 2016-04-07
remove delete node root

alavieille:
[OO-FEATURE] Adding voter to forbid removing node root

### DIFF
--- a/BackofficeBundle/Resources/public/coffee/groupTree/NodeTreeElementView.coffee
+++ b/BackofficeBundle/Resources/public/coffee/groupTree/NodeTreeElementView.coffee
@@ -15,8 +15,15 @@ NodeTreeElementView = OrchestraView.extend(
     nodeGroupRoles = @options.group.model_roles.filter (element) ->
       element.model_id == nodeId
     formCollectionViewClass = appConfigurationView.getConfiguration('group_tab_node_tree_form', 'editEntityTab')
+    roles = @options.roles.roles
+
+    if nodeId == 'root'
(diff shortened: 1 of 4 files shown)


BackofficeBundle/Resources/public/coffee/groupTree/NodeTreeElementView.coffee
```diff
@@ -15,8 +15,15 @@ NodeTreeElementView = OrchestraView.extend(
     nodeGroupRoles = @options.group.model_roles.filter (element) ->
       element.model_id == nodeId
     formCollectionViewClass = appConfigurationView.getConfiguration('group_tab_node_tree_form', 'editEntityTab')
+    roles = @options.roles.roles
+
+    if nodeId == 'root'
+      # A root node can't be deleted
+      roles = @options.roles.roles.filter (element) ->
+        element.name != "ROLE_ACCESS_DELETE_NODE"
```
itkg-ngilain: We should think about a js file containing this constant: the risk is to put it on different location, and the day when it will be changed on the server side, the client side will be hard to maintain :/

itkg-ngilain:
:+1: with or without the constant problem refacto

BackofficeBundle/Resources/public/coffee/groupTree/NodeTreeElementView.coffee
alavieille: I'm agree but we not use backbone model so where put the constant

PR #1564, opened 2016-04-13
fix transform string value type double

alavieille:
[OO-BUGFIX] Fix transform correctly a type double in a string value.


itkg-ngilain:
:+1:

PR #1566, opened 2016-04-13
Change test node with new roles order

KevinArtus:
Change test with new roles order.

https://github.com/open-orchestra/open-orchestra-model-bundle/pull/548
https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/pull/95
.https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1566


alavieille:
:+1:

PR #1570, opened 2016-04-15
fix cloned save button

djamnazi:
If a page contains a cloned button we can not click this button after open a modal which contains a cloned button because the page loses focus. this PR fix that by using two instances: one for normal page and an other for modal page

https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/199


alavieille:
:+1:

PR #1571, opened 2016-04-15
Fix test with add voter workflow

alavieille:
https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/pull/97/files
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1571


itkg-ngilain:
:+1:

PR #1580, opened 2016-04-20
fix test unit command

alavieille:
https://github.com/open-orchestra/open-orchestra-base-bundle/pull/80
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1580


KevinArtus:
:+1:

PR #1585, opened 2016-04-21
add label option into form

KevinArtus:
[OO-TAG] Add label option into FormType


alavieille:
:+1:  but can you be more explicit in OO-TAG what form is updated ?

PR #1589, opened 2016-04-21
Rename left menu

KevinArtus:
[OO-MISC] Rename navigation menu status and Role

https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1589
https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/pull/102


alavieille:
:+1:

PR #1590, opened 2016-04-21
fix update status role update other type node

alavieille:
[OO-BUGFIX] Fix bug updating status of node with the update role of node of other type


itkg-ngilain:
:+1:

PR #1591, opened 2016-04-21
refresh node when submit edit form

alavieille:
[OO-BUGFIX] Refresh node view when edit form of node  is submitted


KevinArtus:
:+1:

PR #1592, opened 2016-04-21
fix error access denied when select other version node

alavieille:
[OO-BUGFIX] Fix error access denied when select an other version node


itkg-ngilain:
:+1:

PR #1593, opened 2016-04-22
fix access denied when edit node without role_access_site

alavieille:
[OO-BUGFIX] Fix access denied when edit node without ``role_access_site``


KevinArtus:
:+1:

itkg-ngilain:
:+1:

PR #1595, opened 2016-04-22
Add user folder create

djamnazi:
[OO-BUGFIX] create user with media folder access and create role for functional test

related PR
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/208


itkg-ngilain:
:+1: if functionnal tests are ok on the back application

PR #1601, opened 2016-04-26
save in textarea on change event tynmce

alavieille:
[OO-BUGFIX] When you save multiple tinymce, there are saved correctly


itkg-ngilain:
:+1:

PR #1602, opened 2016-04-26
fix reset default value when change field type

alavieille:
[OO-BUGFIX] Fix error on field default value if field type is changed in content type form


KevinArtus:
:+1:

PR #1612, opened 2016-04-28
fix global block edition granted

itkg-nanne:
[OO-BUGFIX] Fix block edition right management in global page

### DIFF
--- a/BackofficeBundle/Controller/BlockController.php
+++ b/BackofficeBundle/Controller/BlockController.php
@@ -9,6 +9,7 @@ use OpenOrchestra\ModelInterface\NodeEvents;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration as Config;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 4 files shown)


BackofficeBundle/Controller/BlockController.php
```diff
@@ -41,6 +42,7 @@ public function formAction(Request $request, $nodeId, $blockNumber = 0)
 
         if ($node) {
             $editionRole = $node->getNodeType() === NodeInterface::TYPE_ERROR? TreeNodesPanelStrategy::ROLE_ACCESS_UPDATE_ERROR_NODE:TreeNodesPanelStrategy::ROLE_ACCESS_UPDATE_NODE;
+            $editionRole = $node->getNodeType() === NodeInterface::TYPE_TRANSVERSE? TransverseNodePanelStrategy::ROLE_ACCESS_UPDATE_GENERAL_NODE:$editionRole;
```
alavieille: i think is more explicit create a protected method getEditionRole with if inside like in Node Controller

itkg-nanne:
update done, @alavieille

alavieille:
:+1:

PR #1613, opened 2016-04-28
fix default field in content type form

itkg-nanne:


alavieille:
:+1:  think cherry pick in branch 1.1

PR #1627, opened 2016-05-02
Refresh page when edit website 1.1

alavieille:
[OO-BUGFIX] Refresh page when saving a website form after a form error


itkg-nanne:
:+1:

PR #1634, opened 2016-05-02
redefine right to block controller

itkg-nanne:


alavieille:
i think isn't a good idea to use annotation is granted, for instance if an user has role ROLE_ACCESS_ERROR_NODE but not ROLE_ACCESS_TREE_NODE it can access the action if node is a default node.

itkg-nanne:
update done @alavieille

alavieille:
:+1:

PR #1635, opened 2016-05-02
redefine right to block controller

itkg-nanne:


itkg-nanne:
update done @alavieille

alavieille:
:+1:

PR #1637, opened 2016-05-04
fix show breadcrumb refrest page with datatable

alavieille:
[OO-BUFGIX] Show breadcrumb after refresh page with datatable


djamnazi:
:+1:

PR #1643, opened 2016-05-06
fix pagintation datatable pages undefined

alavieille:
[OO-BUGFIX] Fix the javascript error on datatable pagination


itkg-nanne:
:+1:

PR #1645, opened 2016-05-09
move RoleUsage compiler pass in user admin bundle

alavieille:
https://github.com/open-orchestra/open-orchestra-user-bundle/pull/102
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1645


itkg-ngilain:
:+1:

PR #1654, opened 2016-05-11
fix bo label

alavieille:
[OO-BUGFIX] Add boLabel on root node when create a new website

https://github.com/open-orchestra/open-orchestra-model-interface/pull/183
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1654


itkg-nanne:
:+1:

PR #1660, opened 2016-05-11
Catch js error

itkg-nanne:
[OO-FEATURE] catch js error on form submission


BackofficeBundle/Resources/translations/messages.en.yml
```diff
@@ -38,6 +38,8 @@ open_orchestra_backoffice:
         swchoff:
             on: YES
             off: NO
+        javascript:
+          error: An error occurs during form submission!
```
itkg-ngilain: I think we should find a kind of sentence saying that it's not a server error but a client one

BackofficeBundle/Controller/AdminController.php
```diff
@@ -21,8 +22,12 @@ class AdminController extends Controller
      *
      * @return Response
      */
-    public function adminAction($siteId = null)
+    public function adminAction(Request $request, $siteId = null)
     {
+        if ($request->getMethod() == 'POST') {
+            $this->get('session')->getFlashBag()->add('danger', $this->get('translator')->trans('open_orchestra_backoffice.form.javascript.error'));
```
alavieille: why a flashbag ? and not redirect to backbone view ?

BackofficeBundle/Resources/views/BackOffice/Include/mainPanel.html.twig
```diff
@@ -2,10 +2,13 @@
 <div id="main" role="main">
 
     {% include 'OpenOrchestraBackofficeBundle:BackOffice/Include/MainPanel:ribbon.html.twig' %}
-
+	{% set alert %}
+	    {% include 'BraincraftedBootstrapBundle::flash.html.twig' %}
+	{% endset %}
```
itkg-ngilain: It's a good idea to add the flashbag rendering in this part of the dom, but can you do it via a js mechanism ? I mean that way it works only (if I'm not mistaken) when the entire page is refreshed. But if it was a js mechanism, we could use it anytime we need it without refreshing the page.

itkg-nanne: @itkg-ngilain , we are in a context of js break and form submit with whole page refresh.

BackofficeBundle/Controller/AdminController.php
itkg-nanne: @alavieille , we are in context of js break and form submission. So entire page must be reloaded. I see two ways to do this, use home with a particular message created by flashbag, or a dedicated full page tpl. I choose the first one.

itkg-ngilain: As discussed, I think a dedicated biew is better

BackofficeBundle/Resources/views/BackOffice/Include/mainPanel.html.twig
itkg-ngilain: As discussed, this is not as simple as thought, so let's forget this

itkg-nanne:
update done @itkg-ngilain , @alavieille

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
```diff
@@ -227,6 +227,53 @@
...
+        {% endif %}
+
+        {% if attr.role is not defined or attr.role is empty %}
+            {% set attr = attr|merge({ 'role': 'form' }) %}
+        {% endif %}
+
+        <form  name="{{ form.vars.name }}" method="{{ form_method|lower }}" data-action="{{ action }}" action="{{ path('homepage') }}#error"{% for attrname, attrvalue in attr %} {{ attrname }}="{{ attrvalue }}"{% endfor %}{% if multipart %} enctype="multipart/form-data"{% endif %}>
+        {% if form_method != method %}
+            <input type="hidden" name="_method" value="{{ method }}" />
+        {% endif %}
```
itkg-ngilain: Wasn't it possible to use the parent function to update the form_start instead of copy/paste it ?

itkg-nanne: I don't think so. To do this, I should rewrite the var which is not allowed. @itkg-ngilain

itkg-ngilain:
:+1:

BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
itkg-ngilain: ok

itkg-nanne:
update done @itkg-ngilain , @alavieille

alavieille:
:+1:

PR #1667, opened 2016-05-18
use contentId instead id

alavieille:
[OO-BUGFIX] Fix use contentId instead id in ContentSearchSubscriber

https://github.com/open-orchestra/open-orchestra-display-bundle/pull/216
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1667


itkg-ngilain:
:+1:

PR #1669, opened 2016-05-18
fix content type form for content type

djamnazi:
[OO-BUGFIX] fix contentTypeForm to allow field type modification


itkg-nanne:
:+1:

PR #1673, opened 2016-05-19
fix load tempalte underscore wiithout session

alavieille:
[OO-BUGFIX] Fix loop redirection when load template underscore without valid session


itkg-ngilain:
:+1:

PR #1674, opened 2016-05-19
Fix the content search form type when required

itkg-ngilain:
[OO-BUGFIX] A form containing a required oo_content_search now correctly display an error message when trying to submit it without completing the oo_content_search


itkg-nanne:
TokenInput change the css of the input text to display none.
Maybe you could use the same logic as in OpenOrchestra.FormBehavior.ValidateHidden to remove display:none and add focusable css class.

itkg-ngilain:
@itkg-nanne : I've updated the way you wanted

itkg-nanne:
:+1:

PR #1689, opened 2016-05-25
fix node version voter

alavieille:
[OO-BUGFIX] Fix Backoffice/Security/Authorization/Voter/NodeVersionVoter when create a new node


itkg-ngilain:
:+1:

PR #1691, opened 2016-05-25
1 1 fix content search form type

itkg-ngilain:
[OO-BUGFIX] A form containing a required oo_content_search now correctly display an error message when trying to submit it without completing the oo_content_search


alavieille:
:+1:

PR #1693, opened 2016-05-25
fix configurable content submit and refresh

alavieille:
[OO-BUGFIX] Fix refresh and submit configurable content block form


itkg-nanne:
:+1:

PR #1696, opened 2016-05-25
fix configurable content submit and refresh

alavieille:
[OO-BUGFIX] Fix refresh and submit configurable content block form

cherry-pick https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1693


itkg-nanne:
:+1:

PR #1697, opened 2016-05-25
1 1 fix language tabs

itkg-ngilain:
[OO-BUGFIX] On the node edition page, the node is no more reloaded when clicking on the currently selected language tab


alavieille:
:+1:

PR #1703, opened 2016-05-27
fix duplicate model group role when update group

alavieille:
[OO-BUGFIX] Fix duplicate model group role when update on model group role


itkg-ngilain:
:+1:

Issue #1690, opened 2016-05-25
Add button for content is displayed but create content is not permitted in group


glenouvel:
I create a group that only has access to content.
I affect this group to my user who only this group.
Connect with this user, I see the content list however the add button is displayedbut create content is not active in this group.
Logically when I click it, an 403 error pop

I think the add button must be not displayed

In OO 1.1.1

alavieille:
This is fixed in https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1716 and https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1735. thanks

PR #1716, opened 2016-06-01
hide add button if user d'ont have create content role

djamnazi:
[OO-BUGFIX] hide add button if user d'ont have create content role

https://github.com/open-orchestra/open-orchestra-cms-bundle/issues/1690

### DIFF
--- a/ApiBundle/Transformer/ContentCollectionTransformer.php
+++ b/ApiBundle/Transformer/ContentCollectionTransformer.php
@@ -3,6 +3,7 @@
 namespace OpenOrchestra\ApiBundle\Transformer;
 
 use Doctrine\Common\Collections\Collection;
+use OpenOrchestra\Backoffice\NavigationPanel\Strategies\ContentTypeForContentPanelStrategy;
 use OpenOrchestra\BaseApi\Facade\FacadeInterface;
 use OpenOrchestra\BaseApi\Transformer\AbstractSecurityCheckerAwareTransformer;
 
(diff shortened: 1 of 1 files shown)


ApiBundle/Transformer/ContentCollectionTransformer.php
```diff
@@ -30,7 +31,7 @@ public function transform($contentCollection, $contentType = null)
             array()
         ));
 
-        if ($contentType) {
+        if ($contentType && $this->authorizationChecker->isGranted(ContentTypeForContentPanelStrategy::ROLE_ACCESS_CREATE_CONTENT_TYPE_FOR_CONTENT)) {
```
itkg-ngilain: I'm not sure you check the correct right: I thought the issue was on the content creation. The ROLE_ACCESS_CREATE_CONTENT_TYPE_FOR_CONTENT should be used to allow the creation of content type, not content

djamnazi: @itkg-ngilain  i check the correct right, this is the same right used for the NEW ACTION in contentController. The right for create content type is: ROLE_ACCESS_CREATE_CONTENT_TYPE. you can verify this in the contentTypeController.

itkg-ngilain: @djamnazi : allright so, the naming is ambiguous, but it's not due to your PR.

itkg-ngilain:
:+1:

PR #1717, opened 2016-06-01
1 1 fix move block

itkg-ngilain:
[OO-BUGFIX] a block only used in a single place can now be moved into another area without being deleted

### DIFF
--- a/ApiBundle/Controller/AreaController.php
+++ b/ApiBundle/Controller/AreaController.php
@@ -26,6 +26,9 @@ use OpenOrchestra\ApiBundle\Facade\AreaFacade;
  */
 class AreaController extends BaseController
 {
+    const ROLE_ACCESS = 'access';
+    const ROLE_EDIT = 'edit';
+
     /**
      * @param string $areaId
      * @param string $nodeId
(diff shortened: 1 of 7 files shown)


ApiBundle/Controller/AreaController.php
```diff
@@ -76,23 +77,60 @@ public function showInTemplateAction($areaId, $templateId)
...
+
+    /**
+     * @param string     $nodeId
+     * @param AreaFacade $facade
+     */
+    protected function updateArea($nodeId, AreaFacade $facade)
+    {
+        $nodeRepository = $this->get('open_orchestra_model.repository.node');
+        $node = $nodeRepository->find($nodeId);
+        $this->denyAccessUnlessGranted($this->getAccessRole($node), $node);
```
itkg-nanne: This is the same code as in showInNodeAction. You could maybe factorize it.;

BackofficeBundle/Resources/public/coffee/page/AreaView.coffee
```diff
@@ -68,29 +68,63 @@ AreaView = OrchestraView.extend(
...
+    event.stopPropagation()
+    areaFrom = $(event.currentTarget)
+    areaTo = $(ui.item.context).parent()
+    if (areaTo.find('.newly-inserted').length)
+      @addBlockToArea areaTo
+      return
+    if (ui.sender)
+      @moveBlockToNewArea ui.sender, areaTo
+    else if (areaFrom[0] == areaTo[0])
+      @moveBlockInSameArea areaFrom
```
itkg-nanne: If ui.sender is undefined and areaFrom[0] != areaTo[0] in which case are we?
Could you do :
areaFrom = ui.sender || $(event.currentTarget)
if (areaFrom[0] == areaTo[0])
    @moveBlockInSameArea areaFrom
else
    @moveBlockToNewArea areaFrom, areaTo

itkg-ngilain: @itkg-nanne : This is a 'hard to see' case, you're right. 
When a block move from one area to another, two events are fired: one telling, the block is leaving area1, and one saying the block is entering area2.

The case (ui.sender is undefined and areaFrom[0] != areaTo[0]) is the case when a block leave an area to enter another one. We want to do nothing here as the move is managed l79 when a block enter a zone after leaving another. It's why this case is avoid in the conditions.
With your code the update will be made twice.

itkg-ngilain:
@itkg-nanne : factorization done

itkg-nanne:
:+1:

PR #1719, opened 2016-06-02
Normalize oo_node_choice attr/class option.

itkg-canne:
This piece of code ensure that even if someone sets the `attr` option of a `oo_node_choice` field when adding it inside a form, the html class `orchestra-node-choice` will still be defined.
This class is used to target the field and trigger execution of coffee function `OpenOrchestra.FormBehavior.NodeChoice.activateBehaviorOnElements()`.

### DIFF
--- a/Backoffice/Form/Type/Component/NodeChoiceType.php
+++ b/Backoffice/Form/Type/Component/NodeChoiceType.php
@@ -3,10 +3,11 @@
 namespace OpenOrchestra\Backoffice\Form\Type\Component;
 
 use OpenOrchestra\BaseBundle\Context\CurrentSiteIdInterface;
+use OpenOrchestra\DisplayBundle\Manager\TreeManager;
 use OpenOrchestra\ModelInterface\Repository\NodeRepositoryInterface;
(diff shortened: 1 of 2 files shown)


KevinArtus:
:+1:

itkg-mraiffe:
:+1:

Backoffice/Form/Type/Component/NodeChoiceType.php
```diff
@@ -37,11 +38,30 @@ public function configureOptions(OptionsResolver $resolver)
...
+            if (array_key_exists('class', $value) && !empty($value['class'])) {
+                $classList = explode(' ', $value['class']);
+            }
+            if (!in_array('orchestra-node-choice', $classList)) {
+                $classList[] = 'orchestra-node-choice';
+            }
+            $value['class'] = implode(' ', $classList);
+
+            return $value;
+        });
```
itkg-nanne: You coud use the pattern on new_attr as in https://github.com/open-orchestra/open-orchestra-cms-bundle/blob/master/Backoffice/Form/Type/Component/KeywordsChoiceType.php

PR #1722, opened 2016-06-02
Fix misleading workflow role translations

itkg-ngilain:
[OO-BUGFIX] Fix wrong translation on workflow rights in group administration
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1722
https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/pull/115


itkg-nanne:
:+1:

PR #1723, opened 2016-06-02
Fix broken css on some input text

itkg-ngilain:
[OO-BUGFIX] Fix some broken input


itkg-nanne:
:+1:

KevinArtus:
@itkg-ngilain It's fixed when we wanna add somehting like worflow but there is always the problem when we wanna edit one.

KevinArtus:
@itkg-ngilain We have the problem. 
If you have only one language on your application, you will can get the same problem when you will wanna edit a workflow profil. But, we do not have the problem if we wanna add a workflow profil.

The solution is, if we add a second language on our config.yml, the problem is solve.

PR #1737, opened 2016-06-06
Clear the node per node rights when changing the binding between a group and a site

itkg-ngilain:
[OO-BUGFIX] Clear the node per node rights when changing the binding between a group and a site


itkg-nanne:
:+1:

PR #1738, opened 2016-06-06
Use the site binded to the group when setting a node role instead of the currentSite

itkg-ngilain:
[OO-BUGFIX] Use the site binded to the group when setting a node role instead of the currently selected site
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1738
https://github.com/open-orchestra/open-orchestra-model-interface/pull/191
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/591


itkg-nanne:
:+1:

PR #1752, opened 2016-06-10
control language before creating corresponding node

itkg-nanne:
[OO-BUGFIX] Disallow creation of node in a language belonging fronts languages but not current site aliases languages


alavieille:
:+1:

PR #1753, opened 2016-06-10
exclude removed node from group update action

itkg-nanne:
[OO-BUGFIX] Fix exception on node role list update in group when node contains deleted


itkg-nanne:
update done @alavieille

alavieille:
:+1:

PR #1754, opened 2016-06-10
exclude removed node from group update action

itkg-nanne:
[OO-BUGFIX] Fix exception on node role list update in group when node contains deleted


GroupBundle/EventSubscriber/UpdateNodeGroupRoleMoveNodeSubscriber.php
```diff
@@ -63,46 +63,48 @@ public function updateAccessNodeGroupRole(NodeEvent $event)
...
+                    foreach ($nodeRole as $role => $translation) {
+                        $nodeGroupRole = $group->getModelGroupRoleByTypeAndIdAndRole(
                             NodeInterface::GROUP_ROLE_TYPE,
-                            $node->getParentId(),
+                            $node->getNodeId(),
                             $role
                         );
 
-                        if (null === $nodeGroupRoleParent) {
+                        if (null === $nodeGroupRole) {
```
alavieille: i dont understand this part why throw an exepction if nodeGroupRole is null ?

alavieille:
can you update test for this case ?

PR #1758, opened 2016-06-15
fix popin if error when click on button new version

alavieille:
[OO-BUGFIX] Fix popin error when click on button duplicate version


itkg-ngilain:
:+1:

PR #1762, opened 2016-06-16
Create the OrchestraViewConfigurator before dom:ready

juchi:
Currently when overriding an OrchestraView object (using the appConfigurationView's configuration) we also need to register a specific route for the case of a refresh of the page. This is due to the fact that the appConfigurationView is not registered before the dom:ready event and that the backbone router is started too early in the process (so my custom configuration is not yet registered when the router tries to match a route).

This change fixes this issue and has no functionnal impact on the application since it's only about JS object creation.

### DIFF
--- a/GruntTasks/Options/concat.orchestrajs.js
+++ b/GruntTasks/Options/concat.orchestrajs.js
@@ -6,7 +6,6 @@ module.exports = {
         'web/built/openorchestrabackoffice/js/tinyMCE/activateTinyMCE.js',
         'web/built/openorchestrabackoffice/js/tinyMCE/bbcode2htmlConfigurator.js',
         'web/built/openorchestrabackoffice/js/tinyMCE/html2bbcodeConfigurator.js',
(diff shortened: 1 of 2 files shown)


alavieille:
:+1:

GruntTasks/Options/concat.orchestrajs.js
```diff
@@ -91,6 +90,8 @@ module.exports = {
         'web/built/openorchestrabackoffice/js/table/TableviewCollectionView.js',
         'web/built/openorchestrabackoffice/js/table/tableviewLoader.js',
 
+        'web/built/openorchestrabackoffice/js/viewConfigurator.js',
```
ngilain: @juchi : Can you add a comment code in the same format as other sections please ?

ngilain:
:+1: when a section comment is added

GruntTasks/Options/concat.orchestrajs.js
juchi: updated

juchi:
Updated @ngilain

PR #1763, opened 2016-06-16
fix norm data when submit content form

alavieille:
[OO-BUGFIX] Fix data normalization when submitting a content form


itkg-ngilain:
:+1:

PR #1764, opened 2016-06-17
Fix merge form block strategy

alavieille:
[OO-DEPRECATED] AbstractModulableTypeSubscriber class is deprecated since version 1.2.0 and will be removed in 1.3.0
[OO-DEPRECATED] BlockTypeSubscriber class is deprecated since version 1.2.0 and will be removed in 1.3.0.  use BlockFormTypeSubscriber
[OO-DEPRECATED] GenerateFormManager:createForm method is deprecated since version 1.2.0 and will be removed in 1.3.0.  use getFormType
[OO-BUGFIX] Fix merge form block strategy with transformer and subscriber


itkg-ngilain:
:+1:

PR #1767, opened 2016-06-20
fix norm data when submit content form

alavieille:
[OO-BUGFIX] Fix data normalization when submitting a content form
cherry-pick https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1763


itkg-ngilain:
:+1:

PR #1770, opened 2016-06-21
fix display translated value

djamnazi:
[OO-BUGFIX] Fix the display of active tab of translated value form

https://github.com/open-orchestra/open-orchestra-workflow-function-bundle/issues/114

PR #1772, opened 2016-06-21
fix display translated value

djamnazi:
[OO-BUGFIX] Fix the display of active tab of translated value form

cherry-pick https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1770


alavieille:
:+1:

PR #1783, opened 2016-06-24
Fix transform keywords

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1783
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/608

### DIFF
--- a/BackofficeBundle/Resources/public/less/smartadminPatches/form.less
+++ b/BackofficeBundle/Resources/public/less/smartadminPatches/form.less
@@ -88,7 +88,7 @@ label.required:after {
 }
 
 i.helper-block {
-  position: absolute;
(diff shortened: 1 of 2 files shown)


BackofficeBundle/Resources/public/less/smartadminPatches/form.less
```diff
@@ -88,7 +88,7 @@ label.required:after {
 }
 
 i.helper-block {
-  position: absolute;
+  posittestFindByKeywordsion: absolute;
```
ngilain: @alavieille : I think there a copy-insertpaste-bug here no ?

alavieille:
@ngilain pr updated

ngilain:
:+1:

PR #1788, opened 2016-06-27
 fix category item template in right form

alavieille:
[OO-BUG] Fixing of category of item template in the right form


ngilain:
:+1:

PR #1803, opened 2016-07-06
remove dependency of modelBundle in class fieldTypeType

alavieille:
[OO-BUGFIX] Fix dependency of modelBundle in class fieldTypeType


ngilain:
:+1:

PR #1804, opened 2016-07-06
Fix label form

alavieille:
[OO-BUFGIX] Fix fill label block form

### DIFF
--- a/Backoffice/EventSubscriber/BlockFormTypeSubscriber.php
+++ b/Backoffice/EventSubscriber/BlockFormTypeSubscriber.php
@@ -21,6 +21,7 @@ class BlockFormTypeSubscriber implements EventSubscriberInterface
     {
         return array(
             FormEvents::SUBMIT => 'submit',
+            FormEvents::PRE_SET_DATA => 'preSetData',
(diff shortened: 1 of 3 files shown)


Backoffice/EventSubscriber/BlockFormTypeSubscriber.php
```diff
@@ -35,6 +36,20 @@ public function __construct(array $fixedParameters)
...
      * @param FormEvent $event
      */
+    public function preSetData(FormEvent $event)
+    {
+        $form = $event->getForm();
+        $data = $event->getData();
+        $label = $data->getLabel();
+        $blockPosition = $form->getConfig()->getOption('blockPosition');
+        if ('' == $label && null !== $blockPosition) {
+            $data->setLabel($data->getComponent() . ' #'. ($blockPosition +1));
```
ngilain: Instead of coding here the label generation, maybe you could use a dedicated protected method ``generateLabel($data, $blockposition)`` to isolate the logic

ngilain:
:+1: with or without the modification on logic isolation

PR #1806, opened 2016-07-07
fix label form

alavieille:
[OO-BUFGIX] Fix fill label block form

cherry-pick https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1804


ngilain:
:+1:

PR #1821, opened 2016-07-22
fix refresh datable when remove element

alavieille:
[OO-BUGFIX] No redirect to page 1, when an element of a datable is removed


itkg-nanne:
:+1:

PR #1822, opened 2016-07-22
remove bo label in error node

alavieille:
[OO-BUGFIX] Removing boLabel of the error nodes

https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1822
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/620


itkg-nanne:
:+1:

PR #1824, opened 2016-07-25
fix set currently publish to false when duplicate published content

alavieille:
[OO-BUGFIX] Fix currently published when a published content is duplicate


itkg-nanne:
:+1:

PR #1826, opened 2016-07-25
fix set currently publish to false when duplicate published content 

alavieille:
[OO-BUGFIX] Fix currently published when a published content is duplicate
cherry-pick https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1824#issuecomment-234914951


itkg-nanne:
:+1:

PR #1832, opened 2016-07-29
increase ergo for internal link

itkg-nanne:


Backoffice/Form/Type/Component/NodeChoiceType.php
```diff
@@ -47,9 +51,8 @@ public function configureOptions(OptionsResolver $resolver)
     /**
      * @return array
      */
-    protected function getChoices()
+    protected function getChoices($siteId)
```
alavieille: phpdoc

alavieille:
when phpdoc is fixed :+1:

PR #1836, opened 2016-08-04
optimized NodeGroupRoleVoter

alavieille:
[OO-BUGFIX] Fixed slowness when show node tree, Optimized NodeGroupRoleVoter


itkg-nanne:
:+1:

PR #1849, opened 2016-08-24
authorize status update when user is super admin

alavieille:
[OO-BUGFIX]  Authorize status update when user is super admin


ngilain:
:+1:

PR #1876, opened 2016-09-16
fix configuration field type search node

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1877
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1876


ngilain:
:+1:

PR #1883, opened 2016-09-20
DeleteStatusVoter always return false to avoid long facade generation

ngilain:
[OO-HOTFIX] Temporary disable the possibility to delete a status in order to speed up significantly API responses including status


itkg-nanne:
👍 with test

alavieille:
i understand the problem but i don't understand this pull request on a developing branch, why not wait a fix with group context for instance ?

PR #1885, opened 2016-09-22
A content type can no more be deleted if used

ngilain:
[OO-FEATURE] A content type can no more be deleted if it is used
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1885
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/649
https://github.com/open-orchestra/open-orchestra-model-interface/pull/230


itkg-nanne:
👍

PR #1894, opened 2016-09-30
fix block transformer block private

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1894
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/296


itkg-nanne:
:+1:

PR #1904, opened 2016-10-04
add validatio group constraint check password

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/1904
https://github.com/open-orchestra/open-orchestra-user-bundle/pull/125


itkg-nanne:
:+1:

PR #1920, opened 2016-10-05
1.1 DeleteStatusVoter always return false to avoid long facade generation

ngilain:
[OO-HOTFIX] Temporary disable the possibility to delete a status in order to speed up significantly API responses including status


itkg-nanne:
👍

PR #1953, opened 2016-10-12
Change publication date form type from date to oo_date_picker

ngilain:
[OO-FEATURE] In node edition, publishing and unpublishing dates can be selected via a date picker

### DIFF
--- a/Backoffice/Form/Type/NodeType.php
+++ b/Backoffice/Form/Type/NodeType.php
@@ -83,12 +83,12 @@ class NodeType extends AbstractType
                 'choices' => $this->schemeChoices,
                 'label' => 'open_orchestra_backoffice.form.node.scheme'
             ))
-            ->add('publishDate', 'date', array(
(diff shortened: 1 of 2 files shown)


ngilain:
@itkg-nanne : the picker doesn't appear in the form, but the date picker beahavior is correctly activated and no error is raised. Do you have an idea on the problem ?

itkg-nanne:
As there is two oo_date_picker, it could be a conflict between them. Maybe try with only one. @ngilain

ngilain:
@itkg-nanne : this is not the case as it doesn' t work more with only one. And it's correctly working on content type news with 2 date pickers. Plus there is absolutely no JS error.

alavieille:
@ngilain the date picker work, it's just a problem with a z-indx, if you look DOM, you can see at the end the tag <div id='ui-datepicker-div'>...</div>

ngilain:
@alavieille : update done

alavieille:
:+1:

PR #1969, opened 2016-10-31
Fix set default configuration fixed parameter block

alavieille:
[OO-BUGFIX] Fix set fixed parameter defined in default configuration when a block is created

### DIFF
--- a/ApiBundle/Transformer/BlockTransformer.php
+++ b/ApiBundle/Transformer/BlockTransformer.php
@@ -2,6 +2,7 @@
 
 namespace OpenOrchestra\ApiBundle\Transformer;
 
+use Doctrine\Common\Util\Inflector;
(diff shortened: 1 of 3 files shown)


ApiBundle/Transformer/BlockTransformer.php
```diff
@@ -57,7 +61,8 @@ public function __construct(
         NodeRepositoryInterface $nodeRepository,
         CurrentSiteIdInterface $currentSiteManager,
         TranslatorInterface $translator,
-        EventDispatcherInterface $eventDispatcher
+        EventDispatcherInterface $eventDispatcher,
+        array $fixedParameters
```
ngilain: @alavieille : can you fix the psr2 issue here : `){`

alavieille:
@ngilain updated

PR #2012, opened 2016-12-06
Move workflow fixtures from model to cms

ngilain:
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/695
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2012


alavieille:
:+1:

PR #2029, opened 2016-12-12
Upfate form action

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2029
https://github.com/open-orchestra/open-orchestra/pull/1107

[OO-BCBREAK] Html5 validation on form is removed


ngilain:
:+1: but maybe an OO-tag to indicate there is no more html5 validation ?

PR #2052, opened 2017-01-06
Add accordion

itkg-nanne:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2052
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/715


ApiBundle/Controller/SiteController.php
```diff
@@ -107,6 +107,7 @@ public function listAvailableSiteAction()
     public function deleteAction($siteId)
     {
         $site = $this->get('open_orchestra_model.repository.site')->findOneBySiteId($siteId);
+        $this->denyAccessUnlessGranted(ContributionActionInterface::DELETE, SiteInterface::ENTITY_TYPE);
```
alavieille: the second parameter should be $site not a string for delete action

PR #2072, opened 2017-01-23
fix required template node

alavieille:
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/723
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2072


ngilain:
:+1:

PR #2118, opened 2017-02-24
fix site alias error

alavieille:
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/737
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2118


ngilain:
:+1:

PR #2126, opened 2017-02-27
fix dual alert in case of save and continue for content

itkg-nanne:
https://github.com/open-orchestra/open-orchestra-model-bundle/pull/739
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2126


alavieille:
the problem is also in FormViewButtonsMixin no ? because in _redirectNewElement a FlashMessageBag is create always created

itkg-nanne:
update done @alavieille

ngilain:
:+1:

PR #2141, opened 2017-03-03
Nodes can be moved on leaf nodes

ngilain:
[OO-BUGFIX] Drop on tree now correctly works on leafs
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2141
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/354


alavieille:
:+1:

PR #2143, opened 2017-03-03
Fix broken blocks translations

ngilain:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2143
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/356
https://github.com/open-orchestra/open-orchestra-elastica-bundle/pull/67


alavieille:
:+1: when test is fixed

PR #2145, opened 2017-03-07
Fix s-admin rights

ngilain:
[OO-BUGFIX] fix site admin rights
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2145
https://github.com/open-orchestra/open-orchestra/pull/1161

### DIFF
--- a/ApiBundle/Controller/ContentTypeController.php
+++ b/ApiBundle/Controller/ContentTypeController.php
@@ -37,7 +37,6 @@ class ContentTypeController extends BaseController
      */
     public function showAction($contentTypeId)
     {
-        $this->denyAccessUnlessGranted(ContributionActionInterface::READ, ContentTypeInterface::ENTITY_TYPE);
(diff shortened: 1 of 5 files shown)


ApiBundle/Controller/ContentTypeController.php
```diff
@@ -37,7 +37,6 @@ class ContentTypeController extends BaseController
      */
     public function showAction($contentTypeId)
     {
-        $this->denyAccessUnlessGranted(ContributionActionInterface::READ, ContentTypeInterface::ENTITY_TYPE);
```
alavieille: check if user is authenticated

ApiBundle/Controller/ContentTypeController.php
```diff
@@ -115,7 +114,6 @@ public function deleteContentTypesAction(Request $request)
      */
     public function listForContentAction()
     {
-        $this->denyAccessUnlessGranted(ContributionActionInterface::READ, ContentTypeInterface::ENTITY_TYPE);
```
alavieille: check if user is authenticated

PR #2164, opened 2017-03-16
Fix tree

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2164
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/361


ngilain:
:+1:

PR #2195, opened 2017-03-24
fix highlight menu

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2195
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/368


BackofficeBundle/Resources/public/template/Menu/menuView._tpl.html
```diff
@@ -14,35 +14,35 @@
     <% } %>
     <% if (user.access_section.contribution) { %>
         <li>
-            <a href="#menu-contribution" role="tab" data-toggle="tab" id="navigation-contribution">
```
itkg-nanne: Why remove this? Useless here, i'm ok. But open if an integrator want to highlight just this menu enter.

alavieille: i can use href like a selector

itkg-nanne:
:+1:

BackofficeBundle/Resources/public/template/Menu/menuView._tpl.html
itkg-nanne: href will work as a css class not as an id.

PR #2202, opened 2017-03-28
Fix inject javascript template

alavieille:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2202
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/372


itkg-nanne:
:+1:

PR #2248, opened 2017-04-14
Fix ChoicesOptionToArrayTransformer when data is not of the correct type

ngilain:
[OO-BUGFIX] Fix ChoicesOptionToArrayTransformer when data is not of the correct type


alavieille:
:+1:

PR #2282, opened 2017-04-25
Radio label are placed after button

ngilain:


BackofficeBundle/Resources/views/Form/form_div_layout.html.twig
```diff
@@ -487,12 +487,14 @@
 {% endblock checkbox_row %}
 
 {% block radio_row %}
-    {{ block('form_row') }}
+    {% with {'no_label': true}%}
```
alavieille: you can set label to false

PR #2299, opened 2017-05-04
Redirect after reset password

alavieille:
https://github.com/open-orchestra/open-orchestra-user-bundle/pull/149
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2299


UserAdminBundle/EventSubscriber/ResetPasswordSubscriber.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
ngilain: @alavieille : github is sending an alert on the first char, are you sure of your filde encoding ?

UserAdminBundle/Tests/EventSubscriber/ResetPasswordSubscriberTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
ngilain: @alavieille : alert here too

ngilain:
:+1: when the github alert is identified

PR #2288, opened 2017-05-02
Add content type in content form breadcrumb

ngilain:

PR #2315, opened 2017-05-15
Fix broken content summary route

ngilain:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2288
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2315


alavieille:
:+1:

PR #2336, opened 2017-06-05
fix jquery initialize

itkg-nanne:
https://github.com/open-orchestra/open-orchestra-cms-bundle/pull/2336
https://github.com/open-orchestra/open-orchestra-media-admin-bundle/pull/415

### DIFF
--- a/BackofficeBundle/Resources/public/ecmascript/Lib/jquery.initialize.js
+++ b/BackofficeBundle/Resources/public/ecmascript/Lib/jquery.initialize.js
@@ -79,7 +79,7 @@
     $.fn.initialize = function (selector, callback) {
         let id = $(this).attr('id');
         if (typeof id == 'undefined') {
-            while($("#" + id).length > 0) {
+            while ($("#" + id).length > 0) {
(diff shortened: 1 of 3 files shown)


BackofficeBundle/Resources/public/ecmascript/Lib/jquery.initialize.js
```diff
@@ -70,12 +76,18 @@
     observer.observe(document.documentElement, {childList: true, subtree: true, attributes: true});
 
     // Handle .initialize() calls.
-    $.fn.initialize = function (callback) {
-        msobservers.initialize(this.selector, callback);
+    $.fn.initialize = function (selector, callback) {
+        let id = $(this).attr('id');
+        while(typeof id == 'undefined' || $("#" + id).length > 0) {
```
alavieille: i think the while isn't a good idea.

itkg-nanne: It's the common solution to generate unique DOM id. Have you another idea?

PR #2356, opened 2017-07-26
fix drag & drop zone

itkg-nanne:


BackofficeBundle/Resources/public/less/theme/open-orchestra.less
```diff
@@ -1080,8 +1080,9 @@ table.dataTable {
     }
 }
 
-table.ui-sortable .fa.fa-arrows-v{
-    cursor: pointer;
+.sortable-handle {
```
alavieille: you can reduce scope of this update

